PR #116, opened 2018-09-24
Size fitting of pioneer.

abmpicoli:


ErikSandbraaten:
Maybe we should do a rebase so that we can modify the comment to reference the issue # ?

ErikSandbraaten:
Also, we could consider to review all the pending changes so far in develop to avoid cherry picking commits. Then we could rebase all of them in one go ?

Issue #58, opened 2018-09-18
Autogenerate enums for xml files at compile time


Nightinggale:
I propose using a script I wrote for M:C, which generates enum values to contain the Types from xml.

What it does is it takes a list of strings (xml file names), reads the Types and prints them to enums. This is saved to a temp file and then copied to the header file if the files differs. This way the header file is only touched if something changed which in turn prevents forcing recompilations for no reason at all. Enums are part of the precompiled header file, meaning a change forces all cpp files to be recompiled.

It allows faster performance because the precompiler is aware of values, both for unrolling loops and for stuff like "what is the index for fishing boats?" (the latter took up 1% of the AI CPU time before I cached it). Using enum values supports switch cases, unlike cached values.

The tradeoff from hardcoding is that it will cause bugs if xml files are modified in any way, which affects the enums. Because of this the script inserts ifdef (preprocessor flag), which allows switching between fast hardcoding and relying on xml values, reducing the need for recompiling for people who edit the xml files.

M:C also has autogenerated code containing the md5 checksums of the hardcoded xml files. In hardcoded state, it can detect changes and throw an error to tell people to use the non-hardcoded for modding.

Using an instance of an enum (like YieldTypes) as variables will make the enum string value show up in the debugger instead of the int value. This makes it useful to have enums for debugging even if the compiler doesn't use the enum values.

The script is easy to configure for special rules for files. This is mainly for a few files/enums, which doesn't follow the naming convention. It can also force files to always be hardcoded, like YieldTypes and UnitAI.

The performance of the enum generating script is so good that using the makefile to check dependencies (all the xml files) takes longer than running the script itself. Because of this the script is designed to run each time even if nothing changed.

Semi related: I have code, which allows using ++ as prefix and postfix on instances of enum types. This allows changing:
```cpp
for (int iYield = 0; iYield < NUM_YIELD_TYPES; ++i)
{
      YieldTypes eYield = (YieldTypes)iYield;
```

into

```cpp
for (YieldTypes eYield = FIRST_YIELD; eYield < NUM_YIELD_TYPES; ++eYield)
```

For full support of this, the enum generating script ends the enums with FIRST_(type) = 0. This will help to get rid of typecasting, which is always a good idea for more reasons than just code readability.

The code looks like this (copy pasted from M:C develop branch)

```cpp
// prefix
template <class T>
static inline T& operator++( T& c )
{
	c = (T)(c + 1);
	return c;
}

// postfix
[...]

PR #157, opened 2018-10-02
Makefile will also copy changed files from folder Assets in dev directory to folder Assets in YOURMOD (including files in subfolders).

ShadesOT:


ShadesOT:
> Do you have a realistic use case where YOURMOD is different from the default ..?

Yes. I got my dev mod entirely separated from the civ4col mod folder. As for the reason ... I think I originally wanted to have an untouched latest version of RAR in the MODS folder and a copy eg RARdev for development. But with that changed name I could not easily load savegames from the mod named RAR. ... so I started to copy the changes, which totally contradicted my original intentions. :-D .... (And history and legacy already made their dents again).
So it is realistic, but it is not a usecase. ;-) Thinking about it, makes me want to clean up the situation on my local drive.

Anyways, what is the YOURMOD variable there for then? Let's make this work and then forget about it

Nightinggale:
YOURMOD is a relic of Makefile 1.0 from 2010. It's still in Makefile 2.5 and that makefile aims to be as generic as possible (matches all mods). However we now use a modified version of 2.5 and no longer has to care about generic compatibility. This means we could clean up the makefile.

However I'm not really feeling for touching it. Microsoft's nmake isn't a good makefile tool. If time is to be spend on anything other than quick fixes, rewriting the makefile to use GNU make would be a whole lot better. I have a perl script, which solves #58, but adding it would require everybody to install [Strawberry perl](http://strawberryperl.com/). This however has the bonus that it comes bundled with a bunch of tools normally found in linux, including GNU make.

ShadesOT:
Ok, but is it ok to make these changes?
Btw, which dll is used by the exe, the one in Assets or those in "x plot city radius"?

Nightinggale:
I haven't tested it, but if you have and it works for you, then I don't see why not. If you can benefit from the change, then feel free to make the change.

The exe will always read from the same location as in vanilla and vanilla has the DLL in assets.

ShadesOT:
merge done localy and pushed. commit dc8654ad878590a3a0ba0b3134134dd302e49554 .

Issue #77, opened 2018-09-18
Learning by doing can clear experience


Nightinggale:
There is a feature called "Learning by doing" where units remembers how many turns it has worked in a certain profession. If it has worked enough turns it gets a chance to become professional.

There are a few problems though.

1. it's invisible in the GUI
2. it's an int and it's reset if the profession changes
3. the AI is moving units around in colonies way too often to make use of this

This ticket is about 2. Imagine you have a free colonist set as a fisherman and you want him to become an expert fisherman. After having fished the same plot for 20 turns, a pirate move past, forces the colonist to a new job and then leaves. You can then place the colonist on the water again, but he just lost 20 turns of experience by being moved.

We should also consider what we want with this feature. Should we use the journeyman modcomp? It's basically the same idea, but it creates a trained person, who has +1 production instead of going directly from free colonist to expert.

abmpicoli:
I could swear that I had made an issue "idea" to change the learn by doing code... But can't find it... Probably I've forgot to "save" it... :) 

Option 1: My idea for thatis that units that can "learn by doing" should have a "cache" of professions they have worked previously, and for how many turns... I think a cache of 6 professions is a good balance between performance and functionality. When the unit change to a profession that doesn't exist the cache, the cache entry with the smaller experience goes out.

Option 2: add into the interface a way for the player/AI to lock a profession for "learning by doing". Say , we can lock the colonist to only accept "learn by doing" for the fisherman profession.

Both options 1 and 2 demand changes in the save structure (to save the "lock" or "caches" ).

Option 3: add into the interface/AI a way for specifying to which profession the free colonist wants to become expert. If they change professions, those turns simply wouldn't count for it's expertise... This is save compatible: we simply change how the profession he is learing is filled. If the player/AI doesn't define the profession, then the profession will simply be the first one the colonist is assigned to. We could make the "clear specialty" button to reset the learn by doing training as well..

Option 4: add a "decay" concept. Whenever the player switch professions, instead of reset the old profession, it makes the progress towards expertise be decreased by a specified amount. When the progress reaches 0, the profession to which it is becoming expert changes. This is save compatible, so it is a good option as well. However, it have the downside that training will take much longer...

abmpicoli:
For "Option 1" I've also thought that we could make the rule to be (to have, say,  20 turns on the specific profession or 10 turns at the specific profession and 15 at other professions...)

Nightinggale:
[...]

PR #237, opened 2019-01-05
Partial "fix" to #77 .

abmpicoli:
The increase at GlobalDefinesAlt.xml is increased to 5.

Both CvCity and CvPlayer codes have been changed so that:

1) The minimum turns before a unit can become an expert is adjusted
according to traits that confer a bonus to the LbD .

2) The probability of becoming an expert actually increases at each turn
, as it was the original developer's intention.

3) Experience is no longer set to zero if profession changes, allowing colonists to move to different cities or change professions within the cities without losing experience. It only becomes forbidden for a unit to become an expert that turn.

4) Downside: there is a chance that a unit becomes an expert to a profession it only recently changed into.

### DIFF
--- a/Project Files/DLLSources/CvCity.cpp
+++ b/Project Files/DLLSources/CvCity.cpp
@@ -10832,7 +10832,6 @@ bool CvCity::LbD_try_become_expert(CvUnit* convUnit, int base, int increase, int
 
 	int calculatedChance = (base + (workedRounds - pre_rounds) * increase / l_level);
 	
-
 	if(!isHuman())
 	{
 		int ki_modifier = GC.getLBD_KI_MOD_EXPERT();
--- a/Project Files/DLLSources/CvCity.cpp
+++ b/Project Files/DLLSources/CvCity.cpp
@@ -125,7 +125,7 @@ void CvCity::init(int iID, PlayerTypes eOwner, int iX, int iY, bool bBumpUnits)
 	CvPlot* pAdjacentPlot;
 	CvPlot* pPlot;
 	BuildingTypes eLoopBuilding;
-
+	
 	pPlot = GC.getMapINLINE().plotINLINE(iX, iY);
 
 	//--------------------------------
@@ -236,7 +236,7 @@ void CvCity::init(int iID, PlayerTypes eOwner, int iX, int iY, bool bBumpUnits)
 	{
 		setTeachUnitClass(bestTeachUnitClass());
 	}
-
+	
 	setEverOwned(getOwnerINLINE(), true);
 
 	updateCultureLevel();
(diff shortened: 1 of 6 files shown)


abmpicoli:
> 
> 
> That's quite a lof of whitespace changes ?

You can easily choose to hide whitespaces in the diff...

Nightinggale:
Hiding whitespace changes won't fix the issue with merge conflicts.

abmpicoli:
Done.

abmpicoli:
Unless someone comments otherwise, I intend to merge this branch by Monday ...

Project Files/DLLSources/CvCity.cpp
```diff
@@ -11044,11 +11032,30 @@ void CvCity::doLbD()
...
 	pre_rounds_free = pre_rounds_free * train_percent / 100;
+	for (int iTrait = 0; iTrait < GC.getNumTraitInfos(); ++iTrait)
+	{
+		TraitTypes eTrait = (TraitTypes) iTrait;
+		if (eTrait != NO_TRAIT)
+		{
+			if (hasTrait(eTrait))
+			{
+				pre_rounds_expert *= 100;
+				pre_rounds_expert /= GC.getTraitInfo(eTrait).getLearningByDoingModifier() + 100;
```
Nightinggale: The math is done differently from math elsewhere. Usually it's (100 - modifier)/100. Here it's 100/(modifier + 100). Why this difference?

Another note is to add int iModifer = GC.getTraitInfo(eTrait).getLearningByDoingModifier() and then only do the */ if iModifer != 0. The approach here will result in a lot of *100 /100 operations.

Project Files/DLLSources/CvPlayer.cpp
```diff
@@ -20168,12 +20151,31 @@ void CvPlayer::doLbD()
...
 	pre_rounds_free = pre_rounds_free * train_percent / 100;
+	for (int iTrait = 0; iTrait < GC.getNumTraitInfos(); ++iTrait)
+	{
+		TraitTypes eTrait = (TraitTypes) iTrait;
+		if (eTrait != NO_TRAIT)
+		{
+			if (hasTrait(eTrait))
+			{
+				pre_rounds_expert *= 100;
+				pre_rounds_expert /= GC.getTraitInfo(eTrait).getLearningByDoingModifier() + 100;
```
Nightinggale: same comments as the previous modifier applying code. Also it would likely be a good idea to cache the combined modifier in CvPlayer. This will boost performance and make it easier to make it work with CivEffects later.

Project Files/DLLSources/CvCity.cpp
abmpicoli: Why the value is 100/(modifier+100) : Because the modifier is an increase in rate. If I say learning is 100% faster, I mean that it will take half the time to learn it. If you use (100-modifier)/100, you will end with zero time learning...

Project Files/DLLSources/CvCity.cpp
```diff
@@ -10810,40 +10810,26 @@ bool CvCity::LbD_try_become_expert(CvUnit* convUnit, int base, int increase, int
...
+		return false;		
+	} ;
+	
 
 	if (workedRounds < pre_rounds)
 	{
 		return false;
 	}
-	
+
```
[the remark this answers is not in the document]
abmpicoli: I've tried to "fix" this single space issue , but I don't know why it doesn't appear. But, considering that this code snippet is significant changed as a whole, I doubt this whitespace will be the cause of an issue... right?

Project Files/DLLSources/CvPlayer.cpp
abmpicoli: I'm not sure how do you want me to cache this modifier.... by having an algorithm like this at CvPlayer?
```
CvPlayer::getPreRoundsExpert() {
  if preRoundsExpert == -1 then
      .. //calculate pre Rounds expert 
}
```
?

PR #266, opened 2019-05-19
BUGFIX : Defensive units can now get withdraw promotion

johan-buret:
The original code, inherited from Civ IV, prevented defensive units to get withdrawal promotions .
But we need those for trade naval units at least, for the "Merchant Navy" promotion series.

In RaR or WTP, withdrawal chances represent the ability to flee combat, regardless if the unit is the attacker or the defender. 

I don't think this changes will hurt.


MichaelQuell:
I thought the ships dont get the withdrawal promotions, becuase of the mechanic that if you have built docks in a city there is a chance that a defeated ship will escape, without its goods. Therefore they dont need the withdrawl promotion.

johan-buret:
Escape chances and withdrawal chances are two differents things.

PR #448, opened 2021-05-02
city plot mouse over help

inaiwae:
plot tiles in city show mouse over help (i.e. terrain, features, bonuses, yields, improvements)


raystuttgart:
@inaiwae 
The code looks good.
Will test it ingame.

**However:** 
Please next time, please do not add your **".ini"** to future pull request
But of course small errors like that happen. :)

inaiwae:
sorry for ".ini" file, I never touched it, not sure why it was different and I realized too late it was pushed in as well ... I never worked with github up until a week ago, still learning :)

raystuttgart:
@inaiwae 

No problem considering the .ini file.  👍 
Just wanted to tell you about it, so you know.

I am also not very experienced with GIT - actually a beginner as well.
I have thus also made myselfs wrong commits and needed the help of @Nightinggale to fix.

The important thing is that we notice (e.g. by the reviews) and try to avoid to make the same mistakes again.

raystuttgart:
@inaiwae 
I merged, tested and it is looking fine.
Currently I think it does not need further improvement.

Will be part of the next release and of course you will get credit.
Will also post about your work being integrated in WTP in the forum.

Please consider working again or more closely with us in the future. :)

PR #503, opened 2021-09-23
Fix for memory access violation

MagisterBelisarius:
getUnitInfo() gets uninitialized data since the unit was destroyed previously if any of the other conditions triggered. Thus, getUnitInfo dependent checks must be performed last, not first.


MagisterBelisarius:
Added another proposed fix. Please review as I am unsure of that one.

Issue #426, opened 2021-02-10
move the game screen to the colony when the building availability screen appears


MrZorG33:
discussion on the forum:
https://forums.civfanatics.com/threads/center-of-the-screen-on-the-city-when-an-event-occurs.663321/

raystuttgart:
Finished now.

MrZorG33:
I pasted the code into vDLLButtonPopup.cpp but there is no change. maybe I missed something?

// Ramstormp, WtP, Add Examine Settlement option to the no longer lacking yields for building production popup - START 
		iExamineCityID = 0;
		iExamineCityID = std::max(iExamineCityID, GC.getNumUnitInfos());
		iExamineCityID = std::max(iExamineCityID, GC.getNumBuildingInfos());

		if (pPopupReturn->getButtonClicked() == iExamineCityID)
		{
			CvCity* pCity = GET_PLAYER(GC.getGameINLINE().getActivePlayer()).getCity(info.getData1());
			if (pCity != NULL)
			{
				gDLL->getInterfaceIFace()->selectCity(pCity, true);
			}
		}

		else if (pPopupReturn->getButtonClicked() >= GC.getNumUnitInfos())
// Ramstormp - END

![image](https://user-images.githubusercontent.com/61798883/143787116-3b908e64-aa65-4173-af37-c5d9f2946632.png)

MrZorG33:
I checked - it works.
![image](https://user-images.githubusercontent.com/61798883/143891097-fefe00d8-1416-4a2f-8fd7-a5ef7b2b1a57.png)

but if you enter the city to check the information and leave, then the proposal to build a building disappears and does not appear any more.
![image](https://user-images.githubusercontent.com/61798883/143891387-818680cb-e536-433a-b95f-5e84b544ca21.png)

raystuttgart:
Once you entered the City it is fine for me that the proposal does not appear anymore.
You can make your decision in there or in the Pop-Up the first time.

MrZorG33:
in my opinion this is not correct - we go into the city to see what needs to be built. and if a building is under construction in the city, then we will lose one turn if we manually choose from the proposed one. and if the offer appears again, the building will be built automatically. construction of another building will continue, without delay.
if this is possible, let's try it.

raystuttgart:
Well, if **@Ramstormp** wants to continue working on it, I do not mind. :dunno:
Personally I feel it is fine as it is.

[...]

PR #508, opened 2021-11-06
 move the game screen to the colony when the building availability screen appears #426

Ramstormp:
- Examine city option added (based on my understanding of the solution proposed by raystuttgart)


raystuttgart:
@Ramstormp 
Thanks a lot for this contribution.

PR #517, opened 2021-12-18
Generals Only Promote Army & Admirals Navy

Ramstormp:
Generals and admirals only share experience with units in the army or the navy respectively


raystuttgart:
@Ramstormp 
Thanks for the contribution. :)

Please also check my latest commit - I improved **your feature further**.
I improved the **Unit Selection Pop-Ups** for the Admirals / Generals. 
(Before I did that, the Genral-Popup also showed Civilians even though it did not give them XP anymore.)

But in WTP it now works perfectly.
So do not worry.

PR #518, opened 2021-12-19
Division by zero crash fix

Ramstormp:


raystuttgart:
You actually stumbled on a real bug.  But you fixed it in the wrong place.
(The bug was really really old and very unlikely to trigger -> It may have been the "Negative Yield Bug".

Please check my code now. 
It is now all corrected.

PR #577, opened 2022-02-28
Update CIV4EventInfos.xml

dcrookston:
Fixed a handful of XML errors


raystuttgart:
@dcrookston I took a quick look out of interest and merged since it was no effort. 👍 
Even though currently not modding anymore I can still check from time to time.

PR #650, opened 2022-11-26
Change [Enemy,Neutral,Friendly]HealChange to a value in % of regular heal rate

jo-oe:
as suggested here:
https://forums.civfanatics.com/threads/we-the-people-bug-reporting.636760/post-16374067


jo-oe:
Please note that I had another local change in my repo about the Americas Gigantic map. I have locally reverted it before sending the pull request so it does not interfere. (but it is still visible in the history)

raystuttgart:
I have to refuse this change, because you would be messing with the "positive Heal changes" of the other promotions.
So this change may fix the "negatvie Promotions"; but it would **introduce bugs** in the "positive Promotions".

jo-oe:
I did take care of that though by making the positive promotions into % changes as well, resulting in the same effective change for most situations. Only exception (change to former behaviour, but not bug) would be those situations which are 0 heal without promotion, so there is no possible factor to correct for it.

raystuttgart:
Ok, I checked again and what you did seems to be ok.
(The positive Promotions will also work ok.)

I will merge it. 👍

PR #668, opened 2022-12-18
Fix variable name in CvRandomEventInterface.py

jo-oe:


raystuttgart:
Why are special characters in the comment suddenly strangely encoded?

raystuttgart:
But ok, I will merge it, seems otherwise ok.

jo-oe:
> Why are special characters in the comment suddenly strangely encoded?
That is really weird, seems to be my editor, regardless of the encoding selected, it messes with whitespaces and special chars. I was at least able to fix the many whitespace changes by a git diff -w > somefile ; git revert <rev>; patch -Np1 -i somefile

Nightinggale:
> what is the encoding you are usually using?
CodePage 1252

However it's not the first time I have seen encoding issues. Ideally we should stick to ASCII and English in the source files. It's not catastrophic to have encoding issues in comments through. Not ideal, but not code breaking either.

raystuttgart:
I have just translated these German comments into English - which is better anyways.

raystuttgart:
So was this all about having one variable renamed?

jo-oe:
> So was this all about having one variable renamed?

Yeah. I hate it. Probably I'll have to find a different editor, I have been using atom.io which apparently will stop being developed as of last week. But it had so many cool features ... :(

Nightinggale:
You aren't the only one getting upset that atom.io has been discontinued. The argument is that there are so many better alternatives, yet somehow people can't find them, particularly not when considering the vast library of plugins.

I have never been happy with python editing as I have ended up using notepad++ and as such has no IDE to locate function names or similar. Working with python would be so much easier with a proper IDE.

raystuttgart:
@jo-oe 
**Notepad++** works pretty good and it is what all of us others use.
But yeah, it is not really a programming IDE although it at least does some highlighting well.

PR #669, opened 2022-12-19
Correct typo+incorrect info in Load Game texts

jo-oe:


jo-oe:
This time just using sed - so there is no whitespace / special char damage :)

jo-oe:
Well, reading a little bit more, it might actually rather have been Xicotencatl the Younger, seeing that he opposed the Spanish and Maxixcatzin ... any idea where the quote came from?

raystuttgart:
I merged it. Also it was not about "The Elder" or "The Younger" but about the **title** "Elder" which I corrected.
The rulers were not really "Kings" they were more of "Speakers" / "Elders".

jo-oe:
Well, my information was just a bit of wikipedia-ing and checking documents linked there:
https://en.wikipedia.org/wiki/Xicotencatl
http://www.ejournal.unam.mx/ecn/ecnahuatl32/ECN03204.pdf
Xicotencatl I was almost 100 years old, that's why I thought the quote referenced him. But then he cooperated with the Spanish and made them hang his own son Xicotencatl II. The latter might have been the actual source for the quote - he was critical of the Spanish and an enemy of Maxicatl, but then again Xictotencatl II's actual role seems disputed as there are only accounts from much later and from Spanish sources (which are probably not too objective).

raystuttgart:
It does not matter that much actually. :)
It is just a quote and as long as there are no spelling errors anymore everything is fine.
So let us not waste more time on it than necessary.

PR #670, opened 2022-12-19
Update CIV4GameTextInfos_Original_utf8.xml

ElCid0204:
First try to translate via Github...


ElCid0204:
This was made barely in a simple web browser, but it seems I have to try a GIT Client and Notepad++ combo :-)
Do you have a good practice to make the xml faster? I mean, do I need to insert the >Hungarian<   >/Hungarian<  tags all the time manually or there is a better way to do it? Thanks

ElCid0204:
Thanks for your advice.
I'll work locally, test it and if's OK commit and push. I really don't
wanna mess things up. 🙂

To test I need to download the latest beta of New Hope. Now I need to
figure out to how?

raystuttgart ***@***.***> ezt írta (időpont: 2022. dec. 19.,
H, 19:44):

PR #681, opened 2022-12-29
add -3 yield of cochineal to all terrains except shrublands

jo-oe:


raystuttgart:
Sorry but this is a bad fix.
I have to refuse it.
I already commited the correct fix.

jo-oe:
But is there a reason this is inconsistent with the other yields? I just tried to playtest my fix and it appears to work ...

raystuttgart:
No, the "-" setting for the Terrains should only be used for "Hill Yields" like e.g. Grapes or Red Pepper.
It is not supposed to be used for all Yields generally.

raystuttgart:
Please try my fix.

PR #684, opened 2022-12-30
Fix for path finder so AI can unload their units again

jo-oe:
Please double check this is the correct fix. It was a wild guess ...

### DIFF
--- a/Project Files/DLLSources/KmodPathFinder.cpp
+++ b/Project Files/DLLSources/KmodPathFinder.cpp
@@ -234,6 +234,7 @@ CvPlot* KmodPathFinder::GetPathEndTurnPlot() const
 
 	while (node && node->m_iData2 > 1)
 	{
+		if (node->m_pParent->m_iData2 = 1 && node->m_pParent->m_iData1 > 0) break; // this is the case if we are just 1 plot away from the destination, but will have negative movement points if we move to the destination this turn. We should then return the destination, not our current plot!
 		node = node->m_pParent;
 	}
 	FAssert(node);


Nightinggale:
```cpp
if (node->m_pParent->m_iData2 = 1 && node->m_pParent->m_iData1 > 0) break; // this is the case if we are just 1 plot away from the destination, but will have negative movement points if we move to the destination this turn. We should then return the destination, not our current plot!
```

Just to be safe, wouldn't it make sense to check if parent isn't NULL? If it stops if parent is NULL, then at least it shouldn't crash even if something went wrong.

```cpp
if (node->m_pParent == NULL || (node->m_pParent->m_iData2 = 1 && node->m_pParent->m_iData1 > 0)) break; 
```

jo-oe:
Shouldn't that case at least trigger the assertion above?
In that case wouldn't a crash be better than a silent fail (if assertions are disabled), as there will be crash reports and no weird unexpected behaviour ...

Nightinggale:
What happens now is it loops through nodes and let the following node be node->m_pParent. That works if the parent is NULL as the value is read, but not followed until the next iteration where node will be NULL, hence ending the loop. Your new code will use m_pParent as a pointer without checking if it's NULL and as such it can crash.

```cpp
while (node && node->m_iData2 > 1)
{
#if GLOBAL_DEFINE_USE_CLASSIC_MOVEMENT_SYSTEM == 0
	if (node->m_pParent == NULL || (node->m_pParent->m_iData2 = 1 && node->m_pParent->m_iData1 > 0))
		break;
	// With the new movement system enabled, the end plot is the one m_iData1 reaches 0 (runs out of move) or 
	// m_iData2 has increased, which happens when m_iData1 turns negative and then increases by the movement points for one turn.
#endif
	node = node->m_pParent;
}
```

I think this is more like it. The comment threw me off a bit because it's factually incorrect. The problem also occurs if a unit has 2 moves, moves through a plot with 1 cost and then into a hill meaning it's not just when the target is one plot away. Also let's keep the old code for the classic move system just to be safe.
I don't think m_pParent will ever be NULL, but now we will be sure it won't crash even if it happens.

raystuttgart:
@jo-oe Let us know when you implemented the change as @Nightinggale proposed. 👍

raystuttgart:
**@jo-oe** 
To prevent errors it would be **best** if we **close this Pull Request**.
Once you implemented and tested the suggestion of @Nightinggale please **open a new Pull Request.**

Thanks a lot for trying to fix this.
It is highly appreciated.

PR #694, opened 2023-01-06
fix double definition of Africa trading goods quest finisher

jo-oe:
(and include the other finisher options)
This allows compiling with _HARDCODED again.


raystuttgart:
Let us please have Raubwuerger - the original author -  fix this.
I had already written him about it.

He knows what he wanted to do with this Quest.
Also it needs to e.g. also be checked if the texts and helper texts will then match.

So let us please not do "Quick and Dirty" solutions.
I thus refuse this "Pull Request".

raystuttgart:
Until then, compile a "Dynamic DLL" (Option 2)
It will still e.g. "Hardcode" the "New Movement System".

raubwuerger:
Fixed ...

PR #702, opened 2023-01-09
Fix new movement system for units with 0 max moves (e.g. supplies)

jo-oe:


jo-oe:
note: If I only select the unit with 0 moves, the plots directly adjacent display a red circle (= no movement possible). If I select a group including a unit with 0 moves, the plots directly adjacent display a number of turns (with the fix: MAX_INT) and only the plots in distance >= 2 will display the red circle. I don't know how the interface decided whether to display the number of moves or not, if possible we'd change it so it will also display the red circle for plots in distance = 1 ...

PR #709, opened 2023-01-14
Change some AI logic & make AI a little easier to kill

jo-oe:
This patch will:
1) improve AI load order in Europe so it always loads settlers first
2) fix the endless loop after the last AI settlement has been conquered, where AI transports would loop to Europe to get settlers but fail to load them, travel back to America and then to Europe again.
3) change the number of units for a player to survive from 5 total to 5 excluding treasures, thereby making AI a little easier to kill


raystuttgart:
All changes look good to me.

PR #710, opened 2023-01-14
extend the verifyAlive() checks so wagons, generals and working boats…

jo-oe:
… are also not counted towards the 5 necessary units anymore


raystuttgart:
Generally it is mostly about "How many Units do you still need to hunt down?".
(Otherwise having the logic respawn the Starting Units with a settler.)

So the only stuff that we should exclude there is stuff that is hard to find.

Generals --> easy to find --> so why exclude them specifically ?
Working Boats --> also not harder to find than other ships --> so why exclude them specifically ?
Wagon Trains --> agreed, they can be hard to find if they are in some Native village --> I am fine to exclude them

PR #730, opened 2023-02-12
Update Advisor_utf8.xml

RSP-ARG:
spanish translation


raystuttgart:
There are just 2 lines added? 
Is that correct?

PR #734, opened 2023-02-14
Revision of Player/Team kill logic

jo-oe:
FFs will be recruitable again after Player killed even if they were ignored before


jo-oe:
Please note this includes the change I suggested in discord, where each player is only offered one FF per turn.
(disabling the recursive call CvTeam::testFoundingFather()->convinceFather()->testFoundingFather())

raystuttgart:
**@jo-oe** 
Please contact Nightinggale directly and ask for a review.
(Me may not have seen this since there are so many announcements of changes also he does not visit the forum every day.)

Otherwise you may have to wait for the weekend for me to review it.

PR #832, opened 2023-04-21
Remove yields from import/export list automatically on handleAutoTrad…

jo-oe:
…erouteSetup() if they cannot be transported


raystuttgart:
Ok, I will merge.

PR #840, opened 2023-04-27
Hungarian tranlsation2

ElCid0204:


Schmiddie1979:
I think this is outdated. Hungarian translation already in the making.

Issue #864, opened 2023-05-24
Make a class to access GC, map, game, players etc


Nightinggale:
Right now we have the vanilla system where everything seems to be accessible everywhere. This is horrible for encapsulation as it allows stuff like a const CvPlayer function to use GET_PLAYER to access itself and get non-const access to itself.

We should make an interface class, which has functions like GC, getGame, getMap etc, both const and non-const of each. All classes needing access should then inherit this class, which means any const class function will only have const access to anything else. There will not be any backdoors left where const functions can use GET_PLAYER, GC or similar to gain non-const access to other data or even worse: non-const access to itself.

-----

If the interface functions have their own cpp file, that file can include extern CvGlobals gGlobals and as such skip parts of the call chain as direct access to the instance skips the GC function call. If we take that one step further and make the interface class a friend of CvGlobals, then it can access data like m_map and m_game without extra function calls. Less calls makes stepping through the debugger much cleaner and given how often those functions are called, it has the potential to boost performance too.

f1rpo:
Love this idea, you always use pure virtual classes to good effects. What to name that base class? Seems that all classes that access the game state will derive from it, so maybe "GameStateAccessor". Doesn't exactly sound like something that should exist in well-designed software - but sensible enough under the circumstances.

For a stricter separation between AI and game rule code, there could be one base class that returns CvPlayerAI, CvTeamAI and CvGameAI instances and another that returns CvPlayer, CvTeam and CvGame, but such a separation will require work in other places, in particular, the virtual AI_... functions should then be made inaccessible at the (non-AI) base classes. When e.g. CvCity does need access to an AI player to inform the AI of some game state change, the CvCity instance could be cast to CvCityAI through a wrapper function:
```
// CvCity::init
//GET_PLAYER(ePlayer); // original code
CvPlayerAI& kPlayer = AI().getPlayer(ePlayer);
// (No changes needed below ...)
if (kPlayer.isAlive())
{
	kPlayer.AI_invalidateCloseBordersAttitudeCache();
	aOldAttitude[ePlayer] = kPlayer.AI_getStolenPlotsAttitude(eOwner);
}
```
[...]

Issue #768, opened 2023-03-03
Make GC const to ensure const correctness, thread safe and low OOS code


Nightinggale:
If the code only has const access to CvGlobals, it will only provide a const CvGame reference etc. Most code, which relies on GC will only use it to access functions, which are or should be const.

The way it works right now it is possible to add a const function in CvGame, which then calls GC.getGameINLINE() and that way have non-const access to itself inside the const function. This means that right now we not only lack const correctness, we prevent const correctness by design.

The way to access GC in a non-const way would be to add a non-const GC() to each class, which then provides a non-const reference. This way const functions will not have a way to access non-const GC and through that non-const map, game etc.

An added bonus is that we can use the newly added CvGlobals::isMainThread() to assert that only the main thread gets non-const references. This will make multi-threaded code have read only access to the rest of the code, which will make it harder to make thread-non-safe code. gDLL is a wrapper to GC meaning being const would make exe calls impossible, which again is a win for multi-threaded code as the exe isn't thread safe. We can then write our own gDLL class, which is a wrapper, which can be used in multi-threaded mode. That one then has build in thread locking to ensure thread safe usage.

By making const access the default, the risk of OOS can be reduced and OOS issues should be easier to locate as const functions will become less likely to cause OOS. In fact the only way they should be able to do that would be to generate a return value, which isn't the same on all computers.

We can consider giving GET_PLAYER, GET_TEAM etc the same treatment.

Issue #696, opened 2023-01-06
Make wrapper functions for python interface to info classes


Nightinggale:
Vanilla took a shortcut and provides python direct access to Cv*Info classes (like CvUnitInfo) while all other classes consist of Cy class containing a pointer to a Cv class instance. The use of wrapper functions allow C++ and python to use different types, most notably regarding int vs enum instances.

The exe does call some of the info class functions and expect ints. This can be isolated using wrapper functions and linking redirection.

Solve those two issues and the info classes will only be called by DLL code. This way all the ints can be replaced with the correct types and we can get rid of all the typecasting. Not only will it produce cleaner code, both in code and debugger, but also it will allow the compiler to catch more bugs at compile time. Ideally we should get rid of all typecasting if possible.

f1rpo:
To make such wrapper functions quicker to add, I've written a few macros in my own mod. Usage example for the `int` version, `iPY_WRAP` (not the greatest example because CvProcessInfo doesn't even exist in Civ4Col):
https://github.com/f1rpo/AdvCiv/blob/05f7ce467376cde9ba466f9717346c1623f3883c/CvGameCoreDLL/CvInfo_City.h#L27
Since the wrapper can't have the same name as the DLL-internal function, the function pointer in CyInfoInterface*.cpp needs to be modified too. I'm prepending "py_" to the name:
https://github.com/f1rpo/AdvCiv/blob/05f7ce467376cde9ba466f9717346c1623f3883c/CvGameCoreDLL/CyInfoInterface3.cpp#L139
The actual macros are defined in this header:
https://github.com/f1rpo/AdvCiv/blob/master/CvGameCoreDLL/CyInfoWrapper.h

This does not take care of the (few?) exported functions called by the EXE.

Nightinggale:
> Since the wrapper can't have the same name as the DLL-internal function, the function pointer in CyInfoInterface*.cpp needs to be modified too. I'm prepending "py_" to the name:

I'm wondering about that. Say we add Cy classes to extend the Cv classes, like

```cpp
class CyBonusInfo: public CvBonusInfo
```

Now if the declaration or the implementation of the Cy classes verifies that the size is the same, like this:

```cpp
BOOST_STATIC_ASSERT(sizeof(CvBonusInfo) == sizeof(CyBonusInfo));
```

[...]

PR #870, opened 2023-06-19
Zig Build System & Toolchain

sengir:
This `zig` branch is tracking my efforts to completely overhaul the build system using [Zig](https://ziglang.org/). Zig is an upcoming new programming language, but more importantly, it's also a **build system** _and_ **toolchain** for C, C++, and of course Zig.

## Zig Build System

What makes Zig the build system so compelling is that it lets you **define a static build graph using an imperative language** (in Zig, if you'd believe it), allowing you to easily model the build process without getting lost in a control flow back & forth between variable files, shell scripts, and globbing `nmake` rule. The reason why everyone has stuck to these same old makefiles is because almost no one understands them. Similarly, we could even replace the Perl scripts with Zig scripts to cut our dependence on Strawberry Perl. Then if vendor a portable Zig installation, we could **make the entire mod buildable without a single system dependency**. Pretty cool!

Additionally, the Zig build system integrates seamlessly with an **efficient & robust caching system** of any build artifact, not just intermediate object files. This not only allows us to **shed our reliance on a precompiled header**, but it also lets us **model our codegen build scripts as regular build steps** that depend on specific input files and produce cacheable output files. With this, we should be able to clean up the source code organization quite a bit.

## Zig Toolchain

The real "killer feature" of Zig, though, is its capabilities as a C/C++ toolchain. While the Zig language is capable of natively interfacing with C, `zig cc`/`zig c++` is an alternate mode of the Zig compiler that leverages an embedded Clang and LLD to provide a truly portable C/C++ compiler than can cross-compile from any system to any target system without the usual insanities: cross-compilers, sysroots, etc. (see [here](https://andrewkelley.me/post/zig-cc-powerful-drop-in-replacement-gcc-clang.html) for more info). This functionality is so useful and novel that (Uber uses Zig to compile all of their C/C++ code, allowing them to migrate to arm64)[https://www.uber.com/blog/bootstrapping-ubers-infrastructure-on-arm64-with-zig/]. What this allows us to do is **develop WTP from Linux** (or macOS)! I personally run NixOS and typically run WTP through Steam under Proton, but the current WTP build tooling is incompatible with both Linux and WIne/Proton, requiring either dual booting or running a virtual machine to build the project. Not for long!

The rub here is that this uses Clang to compile our C++ code. Believe it or not, an MSVC++ Toolkit 2003 project is absolutely not compatible with Clang whatsoever. However, I've been hard at work trying to make it compatible! So far, the project builds successfully and I can open up the game and screw around. However, there are many failed `FAssert`s in this build. I believe this is mostly or perhaps even entirely due to latent Undefined Behavior that already existed in the codebase, but MSVC++ Toolkit 2003 was simply not smart enough to optimize that Undefined Behavior into a runtime bug while Clang on the other hand is. Yes, this does mean that this new build of WTP should be **significantly more optimized** thanks to two decades of compiler innovations. We also get two decades of compiler innovations in **static analysis and warnings**, which is perhaps even more beneficial.

Another fun benefit of Clang is that it gives us **C++23** & all of its goodies! Unfortunately, a lot of those goodies are missing since we're still including an STL from MSVC++ 2003. We can still take advantage of the many language-level features, but more interestingly we can use this as an opportunity to define our own sort of STL that caters to our more specific needs and opinionated views. For example, we could make immutability the default that must get opted out of instead of the typical opt-in, allowing us to abide by the principle of least privilege (see #768 #864 #696 etc). In my opinion, the most important gain here would be **cleaning up all the monstrous enum-related metaprogramming** that had to not only be MSVC++ 2003 compatible but also use codegen'd enum stuff _and_ support optionally hardcoding enum lengths. Bravo, @Nightinggale for the hard work, I don't think I could've done it, but holy **** we've got to simplify that ASAP with C++23.

## Next Steps

- [X] Model build graph in `build.zig`
- [X] Get MSVC++ Toolkit 2003 and Windows SDK headers compiling under LLVM (see [WTP/Compiler/zig](https://github.com/We-the-People-civ4col-mod/Compiler/tree/zig))
- [X] Get codebase compiling under LLVM
- [X] Clean up enough Undefined Behavior to get in-game without crashing
- [ ] Continue hunting down Undefined Behavior in the existing C++ code
- [ ] (Optional) Rewrite Perl scripts with Zig scripts
- [ ] Clean up & flesh out `build.zig` build script to accommodate every use case
- [ ] Vendor portable Zig installation in [WTP/Compiler/zig](https://github.com/We-the-People-civ4col-mod/Compiler/tree/zig)
- [ ] Create small batch scripts to simplify basic workflow
- [ ] Rebase this branch on `develop`
- [ ] Beta test this new build system & workflow with WTP developers
  - If you have any issues (e.g. IDE support, broken workflow, anything at all) with the new tooling, let me know and we can work together to figure out a solution.
- [ ] Address WTP developer concerns
- [ ] Upstream to `develop`

## Note

If anyone is interested in playing around with this (or fighting the good fight against Undefined Behavior), then follow [these steps](https://ziglang.org/learn/getting-started/) to install Zig. The build system API got reworked recently but there hasn't been a tagged release since then, so make sure you select the **master branch**/**nightly** build of Zig. In the future, we'll use a vendored copy of Zig, but I'd like to at least wait until the next tagged release before thinking about vendoring anything. For any assistance whatsoever, feel free to message me on here or the Discord server. I'd be happy to help however I can!

NOTE: This is a draft branch; this effort is far from complete. This branch will undergo many changes, including liberal use of history rewriting and force pushes, so keep that in mind if you do checkout this branch.


Project Files/DLLSources/EXE_interface.cpp
ghost, line 1: The name mangling issues could be fixed by passing the `/Zc:wchar_t` compiler flag. Would be much easier, that flag is basically enough to fix the exports for the normal Civ 4 so it shouldn't be much different here.

ghost:
>The rub here is that this uses Clang to compile our C++ code. Believe it or not, an MSVC++ Toolkit 2003 project is absolutely not compatible with Clang whatsoever. 

This is overstated because it's actually surprising how compatible clang has become. The msvc compiler especially from 2015 onwards is less compatible than clang because Microsoft made lots of runtime changes and the compiler relies heavily on the runtime.

You could give https://github.com/backengineering/llvm-msvc a try should you encounter issues with the normal clang compiler.
Edit: This fork probably can't help you because of the way the compiler is integrated in zig.

Project Files/DLLSources/EXE_interface.cpp
ghost: I forgot `_MBCS` must be defined as well.

sengir: We are actually intentionally disabling distinct `wchar_t` via Clang's cc1 flag [`-Xclang -fno-wchar`](https://github.com/We-the-People-civ4col-mod/Mod/pull/870/files#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR348-R350) (AKA MSVC's `/Zc:wchar_t-`). This is to make Boost.Python bindings not break on `wchar_t` being a distinct type from `unsigned short int`. Having to manually remangle `wchar_t` is certainly less than ideal, but I couldn't find a way to get Boost.Python to behave correctly otherwise. I'm not sure if it's because our masquerading as MSVC++ 2003 is confusing Boost or because the EXE's initialization of Boost.Python's converter registry is taking precedence over ours.

If you've got more insight into what exactly the EXE does with Boost.Python, I'd love to hear more. I do know the EXE binds a good number of "secret" types and functions (such as the Scaleform GFC stuff) regardless of what the DLL does.

sengir:
@alberts8:

> This is overstated because it's actually surprising how compatible clang has become. The msvc compiler especially from 2015 onwards is less compatible than clang because Microsoft made lots of runtime changes and the compiler relies heavily on the runtime.

Note: The entire PR description is from an old revision of this work. If you're interested in the current state of this branch, read the [doc comment of `build.zig`](https://github.com/We-the-People-civ4col-mod/Mod/pull/870/files#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR1-R50).

Admittedly, yeah, this is completely hyperbolic. If you use [the right incantation of compatibility flags](https://github.com/We-the-People-civ4col-mod/Mod/pull/870/files#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR334-R346), then it's not too bad. However, there are a good number of very important differences that even [Clang's MSVC compatibility mode](https://clang.llvm.org/docs/MSVCCompatibility.html) fail to achieve. The one that frustrates me the most is of course filesystem case sensitivity (and of course [the open pull request for it is stale](https://reviews.llvm.org/D21113)). There are still a number of small things that are really dumb but you have to discover via extended troubleshooting, such as [missing `uintptr_t`](https://github.com/We-the-People-civ4col-mod/Mod/pull/870/files#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR495), the whole `wchar_t` wackiness mentioned in the other comment, or a _massive_ amount of small discrepancies in how templates work (missing `typename`, missing `template`, missing `this->`, uninstantiated bodies now being parsed, etc.).

> You could give https://github.com/backengineering/llvm-msvc a try should you encounter issues with the normal clang compiler. Edit: This fork probably can't help you because of the way the compiler is integrated in zig.

That's a really cool project, thanks for sharing! You're correct in that it's unsuitable due to how Zig (currently) integrates with LLVM. However, that may change [once LLVM gets removed from the Zig binary (and turned into a build module?)](https://github.com/ziglang/zig/issues/16270). In that case, it may actually make sense to maintain a fork of this hypothetical build module that's based on this llvm-msvc fork.

In the meantime, it seems like I've got things mostly working on par with the current MSVC++ 2003 build. The major remaining issues are due to bugs in the original code that are only being turned into actual problems due to the new compiler not being as a _lenient_ as MSVC. You know, things like obscene UB that should have never compiled in the first place.

If you're curious about any of this, feel free to hop on by the WTP Discord server! I'd love to chat more about this if you are.

Project Files/DLLSources/EXE_interface.cpp
ghost: >Having to manually remangle wchar_t is certainly less than ideal, but I couldn't find a way to get Boost.Python to behave correctly otherwise.

What error is that causing exactly on your side?

Project Files/DLLSources/shim.zig
```diff
@@ -0,0 +1,19 @@
...
+    pDC: ?*anyopaque,
+) callconv(.C) c_int;
+
+export fn __CxxFrameHandler3(
+    pExcept: ?*anyopaque,
+    pRN: ?*anyopaque,
+    pContext: ?*anyopaque,
+    pDC: ?*anyopaque,
+) callconv(.C) c_int {
+    return __CxxFrameHandler(pExcept, pRN, pContext, pDC);
```
ghost, line 14: Calling the `__CxxFrameHandler` instead fixes compilation but the exception handling won't work. The program will terminate if a exception occurs. I'am having this issue as well and there are some options to fix that which I'am evaluating currently.

ghost: This is caused by the magic number having a lower value in the msvc 7.1 runtime. The handler doesn't handle the newer number aka
https://github.com/llvm/llvm-project/blob/0fea00d48bcd2368f21ecbca80833093f3a96e64/llvm/lib/CodeGen/AsmPrinter/WinException.cpp#L735

You could try to reduce that number by 2 when calling the `CxxFrameHandler` maybe that will work. This is one possible solution which i haven't tried myself.

ghost:
Little things like this line of code are something that can be frustrating when using clang.
https://github.com/We-the-People-civ4col-mod/Mod/blob/ba1323a75b478ff61dd8dfafe196d3b530f5d357/Project%20Files/DLLSources/CvGameTextMgr.cpp#L9436

The format specifier and the second argument don't match. This works with msvc but when using clang the resulting formatted string is garbage.

raystuttgart:
This one should be checked by **sneed**.
I will not touch it myself. :)

PR #890, opened 2023-09-30
French translation on achievements

johan-buret:
Updating the French translation that is severely lagging.


raystuttgart:
Ok I checked and only French Text was modified.
Will merge it. 👍

Issue #866, opened 2023-05-30
too much fur


MrZorG33:
in the mid to late game natives start spamming fur. they offer to buy it or give it as a gift in large quantities. other resources are also sometimes present, but against the background of the amount of fur, this is a "drop in the ocean". Is it possible to somehow influence the variety of goods offered by the natives?

IFoundabugg:
Seriously, this has been an issue for a long time. In mid to late game, I am exporting more fur than every other yield combined. As you say, almost every turn the natives offer fur, and if you refuse they gift it soon after in increasingly large quantities.

raystuttgart:
In the current logic of the "Native Trading DLL Diplo Event" they will offer the Yield they have most.
And yes in many cases this will be fur - because from fur they also get Food which makes a lot of sense for them.

It was the logical choice for me that Natives would trade or gift most what they have most.
But if somebody wants to reprogram the logic to be more varied and chose the Yield traded more randomly it of course can be done.

It is simply like always a matter of having somebody motivated to invest his time and effort ...

victorxup:
Pretty simple solution for this: weighted probability depending on how much of each yield stored in native home city.

I've implemented it locally, can make pull request if you are interested.

raystuttgart:
Sure, create a pull request with some comment and explanation. 👍 
Ideally also reference this issue somehow.
We can then read it and will merge it if it looks good.

victorxup:
I suppose, then I need permission to create upstream branch or otherwise a hint - how to do it without access granted))

raystuttgart:
I thought a Pull Request would not need special permissions ...
But I will ask @Nightinggale 

It is however already pretty late here in Central Europe.
So maybe it will take until tomorrow for an answer.

Nightinggale:
> I thought a Pull Request would not need special permissions ...

The whole idea with pull requests is that they can be made without any permissions or connections to a project. Anybody with a GitHub account can make them, possibly for any project. Permissions will not be needed unless somebody is to make a commit, which should be added without being reviewed prior to being added.

MatthewForrester:
> I suppose, then I need permission to create upstream branch or otherwise a hint - how to do it without access granted)

[...]

PR #903, opened 2023-10-12
Issue 866 too much fur 

victorxup:
Proposal to solve Issue [#866] (https://github.com/We-the-People-civ4col-mod/Mod/issues/866)

Possibly need some codestyle enhacement regarding to CPP conventions

### DIFF
--- a/Project Files/DLLSources/CvCity.cpp
+++ b/Project Files/DLLSources/CvCity.cpp
@@ -9275,7 +9275,7 @@ void CvCity::ejectTrader()
 	}
 }
 
-bool CvCity::canProduceYield(YieldTypes eYield)
+bool CvCity::canProduceYield(YieldTypes eYield) const
 {
 	CvPlayer& kPlayer = GET_PLAYER(getOwnerINLINE());
 	for (int i = 0; i < GC.getNumProfessionInfos(); i++)
--- a/Project Files/DLLSources/CvCity.h
+++ b/Project Files/DLLSources/CvCity.h
@@ -749,7 +749,7 @@ public:
 	void ejectTeachUnits();
 	void ejectMissionary();
 	void ejectTrader();
-	bool canProduceYield(YieldTypes eYield);
+	bool canProduceYield(YieldTypes eYield) const;
 
 	bool educateStudent(int iUnitId, UnitTypes eUnit);
 	bool canTeach(UnitTypes eUnit) const;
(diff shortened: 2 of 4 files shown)


devolution79:
Thank you for contributing :)

Some quick style comments: Please make the the member functions const and pass the city as a const reference rather than a pointer.
You can use YieldTypes as the iteration variable rather than an integer (look at similar loops scattered around the code). 
A single random draw would also be preferable. The performance impact is not very important for this rarely called code but the random calls do clutter the random gen. logs and may make it harder to debug OOS.
A comment / description would be helpful to to know that the function does at a glance (i.e. randomly chooses a yield weighted  proportionally by the amount of valid yields available)

devolution79:
isYieldValidForNativeTrade : Could also deny musket and weapon stuff from being traded.

Nightinggale:
First of all the fix is most welcome. It almost feels somewhat ungrateful to point out so many imperfections when it's overall of good quality. However at the same time we do value code quality and consistency to avoid current and future bugs.

> You can use YieldTypes as the iteration variable rather than an integer (look at similar loops scattered around the code).

I added the ability to iterate enums and we are slowly converting all int iterations to enum iterations. The line should look like:

```cpp
for (YieldTypes eYield = FIRST_YIELD; eYield < NUM_YIELD_TYPES; ++eYield)
```

This will remove the need to typecast, which in turn makes debugging nicer and reduces the risk of messing up enum types (bugs).

> A single random draw would also be preferable. The performance impact is not very important for this rarely called code but the random calls do clutter the random gen. logs and may make it harder to debug OOS.

The random call is in the initializer, hence only called once. Other than that I agree. It should be set outside the loop and the resulting value should be const as it's not the iterator. It would make more sense to iterate over a YieldTypes as that makes the code reveal the intention better.

isYieldValidForNativeTrade should check for >= FIRST_YIELD before calling canProduceYield. This way output will be defined for all input and as such reusing the function later and calling it with NO_YIELD will not cause a bug.

We use the coding style where { has a dedicated line. (ok this one could be nitpicking)

EDIT: NUM_CARGO_YIELD_TYPES exist and could be better suited than NUM_YIELD_TYPES. This will exclude looping yields such as bells or hammers, which can't be transported, hence can't be traded.

victorxup:
> isYieldValidForNativeTrade : Could also deny musket and weapon stuff from being traded.

This behaviour was already implemented via getNativeSellPrice() > 0 check, so no need to implement anything new about it.
There's a bunch of goods that natives never gonna sell

victorxup:
Made some changes
Please take a look

Nightinggale:
Now it looks fine. Thanks for your contribution. Feel free to drop by in our discord.

PR #915, opened 2023-10-21
Promotion cache

johan-buret:
The goal of this PR is two fold:

1. Cache the fact wether a promition is available for a defense only unit
2. Forbid the defensive-only ships to go sea fishing, as this promotion gives irrelevant attack bonuses . cf ``CvPromotionInfo::calculateAvailableForDefensiveUnit`` last `for`  loop

This way, the code is a bit cleaner and the calculation is only done once per session.


Nightinggale:
I dislike how the cache is set. It's much better to set once and then read the cache unconditionally. The question is how to do that using clean code?

Right now xml reading is a 4 step process

1. read lengths of xml files
2. read basic info (like Type)
3. read the xml data, now with awareness of all types
4. calls a function to set various caches based on xml data (like which yields are demanded by any unit)

If I add a step between 3 and 4, then that step can call a function in each instance of an info class, let's call it postLoadFixes. I add that function to the base class with no code in it and then each info class can add that function and use it to set up cache. I think that's the cleanest approach, which also supply us with an easy to use reusable code structure.

I might as well get started on this function right away as I know something else, which can make use of it.

-----

Another comment is that you should set it up to loop enum instances instead of ints and set the max like NUM_FEATURE_TYPES instead of calling GC each time. This will make the code appear so much cleaner in a debugger.

johan-buret:
I will close this pull request and come back with the requested changes

PR #916, opened 2023-10-21
Promotion cache

johan-buret:
The ``CvPromotionInfo``  class  now stores as  properties ``isAvailableForDefensiveUnit()`` and ``isNotAvailableForDefensiveUnit()``, who are calculated only once per session.

It now uses the ``postLoadSetup()``  pattern set up by @Nightinggale  earlier today.


Nightinggale:
One tiny detail. The coding standard states iTerrain is an int while eTerrain is an enum. All 3 loops loop enums while using i prefix.

Other than that it seems to be fine now. I haven't tested functionality, but I'm not spotting any flaws there by reading the code so if it passes your ingame tests, then it's likely fine.

johan-buret:
Although complimentary reviews and tests never hurt, it passes mine on these  various cases

- Merchant ship with defense only: Caravela
- Merchant ship with attack capability : Galleon
- War ship : Brig-of-War
- Pirate Ship : Pirate Frigate.

johan-buret:
I've corrected this code to follow the coding style  guideline.

Nightinggale:
In CvCivilizationInfo I added postLoadSetup() right after read(). Since it's a file for all cache of the class, not just this, putting it in between code for this cache isn't the most obvious place to put it long term.

PR #918, opened 2023-10-23
Add map size dependent Ice Caps generation near poles

victorxup:
Add function that creates Ice Caps on poles with slightly random penetration, depending on following factors:
- map size;
- tile surrounding;
- how far from edge of the map tile actually is.

The resulting caps should form irregular shape, but still remain quite solid

### DIFF
--- a/PrivateMaps/FaireWeatherTweakEx.py
+++ b/PrivateMaps/FaireWeatherTweakEx.py
@@ -361,7 +361,11 @@ class MapConstants :
         # Belisarius - Large Rivers
         self.LARGE_RIVER = 11
         self.LAKE = 12
-  
+
+        # We need to set initial default values here. They will be changed when we know actual grid size.
+        self.iceCapMaxHeightNorth = 4
+        self.iceCapMaxHeightSouth = 4
+
         return
     
     def initInGameOptions(self):
@@ -5102,6 +5106,11 @@ def generateWetland():
                                 plot.setTerrainType(terrainWetland, True, True)
                         
 def afterGeneration():
+
+    # Those are scaled depending on map size, and will scale also if new map sizes are introduced
+    mc.iceCapMaxHeightNorth = 2 + int(round(PRand.random() * 0.224 * (mc.height ** 0.34))) + mc.height // 21
+    mc.iceCapMaxHeightSouth = 2 + int(round(PRand.random() * 0.224 * (mc.height ** 0.34))) + mc.height // 21
+
     gc = CyGlobalContext()
     mmap = gc.getMap()
     em.initialize()
@@ -5118,7 +5127,7 @@ def afterGeneration():
                 continue
             if x > mc.width/3 and x < 2 * mc.width/3:#dont penetrate past 1/3 of map
                 continue
-            if y < 4 or y > mc.height - 5:#make room for ice
+            if y < mc.iceCapMaxHeightSouth or y > mc.height - 1 - mc.iceCapMaxHeightNorth: #make room for ice
                 continue
             landFound = False
             for yy in range(y - mc.distanceToEurope,y + mc.distanceToEurope + 1,1):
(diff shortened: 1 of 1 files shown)


johan-buret:
I like the direction this pull is taking, but I have some suggestions

- Python allows to do checks like ``lower_bound <= x < upper_bound``  with the expected mathematical result 
- please explain where the constants you use for your calculation comes from.

victorxup:
> * please explain where the constants you use for your calculation comes from.

It's an empirical approach. I know approximately how function should behave - and then use math methods to reveal the function approximation itself. After several iterations I got the result very close to what I wanted to see.
I tested every map size a bunch of times to ensure that. 
Also I've added comment to the code: Those are magic numbers, empirically balanced.
So feel free to suggest your own balancing if current doesn't provide good results for you.

victorxup:
> * Python allows to do checks like `lower_bound <= x < upper_bound`  with the expected mathematical result

Didn't know that, my knowledge of Python isn't perfect, that's obvious)
Though I still don't see where this can be applicable in my commit.
Please mention the place for it more explicitly

PrivateMaps/FaireWeatherTweakEx.py
```diff
@@ -5161,42 +5170,156 @@ def afterGeneration():
...
+
+    # There is a very little chance that a tile can be ice-capped without corresponding surrounding
+    initialChance = 0.002
+    chanceIncreaseAdjacent = 0.35
+    chanceIncreaseDiagonal = 0.18
+    chanceReductionFromLatitude = 0.20 * mc.height ** 0.287
+
+    numOfIterations = 3
+
+    def isPlotOutOfBounds(x, y):
```
johan-buret: I would replace this function by :
```Python
    def isPlotWithinBounds(x, y):
         return 0 <= x < mc.width and 0<= y < mc.height
    
    def isPlotOutOfBounds(x, y): 
         return not isPlotWithinBounds(x, y)
 ```
 
 All usage of ``isPlotOutOfBounds(x, y)`` I have seen were preceded by a ``not``,  It may be better to use the direct function instead of the double negation.

File: PrivateMaps/FaireWeatherTweakEx.py [hunk 2]
```diff
@@ -361,7 +361,11 @@ def initialize(self):
         # Belisarius - Large Rivers
         self.LARGE_RIVER = 11
         self.LAKE = 12
-  
+
+        # We need to set initial default values here. They will be changed when we know actual grid size.
+        self.iceCapMaxHeightNorth = 4
```
johan-buret, line 366: right call here

PrivateMaps/FaireWeatherTweakEx.py
```diff
@@ -5223,7 +5346,7 @@ def createIce():
 ##cm.printTempMap(cm.summerTempsMap)
 ##cm.printTempMap(cm.winterTempsMap)
 ##cm.printTempMap(cm.averageTempMap)
-            
+
```
johan-buret: Please don't leave meaningless spacing change or add commented code within a functional PR.

PrivateMaps/FaireWeatherTweakEx.py
```diff
@@ -5102,6 +5106,11 @@ def generateWetland():
                                 plot.setTerrainType(terrainWetland, True, True)
                         
 def afterGeneration():
+
+    # Those are scaled depending on map size, and will scale also if new map sizes are introduced
+    mc.iceCapMaxHeightNorth = 2 + int(round(PRand.random() * 0.224 * (mc.height ** 0.34))) + mc.height // 21
```
johan-buret, line 5111: This formula is good enough for what it does. I won't argue on its merit.

PrivateMaps/FaireWeatherTweakEx.py
```diff
@@ -5161,42 +5170,156 @@ def afterGeneration():
...
+    def isPlotOutOfBounds(x, y):
+        if x < 0 or x >= mc.width or y < 0 or y >= mc.height:
+            return True
+        else:
+            return False
+
+    def calculateIceChanceFromAdjacent(x, y):
+        # More chance for Ice if adjacent tiles are Ice, Permafrost or map bounds
+        totalChance = initialChance
+        for direction in range(1, 5):
```
johan-buret: an alternate way to write it in a single loop would be :
```Python
      # this list will give the right chances depending on the directions, as they are well ordered
       chances =[0] + [chanceIncreaseAdjacent] *4 + [chanceIncreaseDiagonal] *4
      
       for direction in range(1, 9):
            xx, yy = GetXYFromDirection(x, y, direction)
            if isPlotWithinBounds(xx, yy) and (xx, yy) in icePlots:
                totalChance = addProbability(totalChance, chances[direction])
       return totalChance
```

PrivateMaps/FaireWeatherTweakEx.py
```diff
@@ -5161,42 +5170,156 @@ def afterGeneration():
...
+            print lineString
+        lineString = " "
+        print lineString
+        return
+
+    # In this dictionary we store Ice plots which should be generated on water and Permafrost tiles on land
+    # (both cause probability increase for neighbouring plots to become ice)
+    icePlots = {}
+    gc = CyGlobalContext()
+    mmap = gc.getMap()
```
johan-buret, line 5255: ``gc``,``mmap`` and stuff like this are defined at a global scope within the map script. I now this is not a good coding practice from the people that came before us, ~~but reassinging them here might not be the best move.~~

I note the previous version of ``createIce`` did that as well. So, yeah, let's live with it.

johan-buret (changes requested):
Just cleanup the white lines and the commented code at the end of the file, and it's good to go.

PrivateMaps/FaireWeatherTweakEx.py
victorxup: Makes sense

PrivateMaps/FaireWeatherTweakEx.py
victorxup: Didn't get what you mean

PrivateMaps/FaireWeatherTweakEx.py
victorxup: Fixed

PrivateMaps/FaireWeatherTweakEx.py
johan-buret: Well. moving the defaults, like you did was an excellent move. I like to point out good ideas too.

PrivateMaps/FaireWeatherTweakEx.py
victorxup: Agree, I tried to avoid wherever possible.
Though:
- I'm just not aware how to do it efficiently in current architecture
- It's not a narrow place when we speak about performance.

Nightinggale: >     * I'm just not aware how to do it efficiently in current architecture
>     * It's not a narrow place when we speak about performance.

Agreed. Improved code design and performance is indeed worthy topics, but comments to a pull request is not the place.

johan-buret (approved):

PrivateMaps/FaireWeatherTweakEx.py
victorxup: Applied your approach

johan-buret (approved):

johan-buret:
All my tests are OK. The polar caps are sturdier and extend further dowwards like they should.
I also tested the case where no North pole cap should be generated. That works too.

PR #930, opened 2023-10-29
Modified faireweather for more realistic desert/barren landscapes

maushacks:
1) It confines the 'hot' sand deserts to around the horse latitudes as in the real world. Or to clarify: The yellow bright sand texture and cactee it can feature don't quite fit the higher up latitudes, hence:
2) It ramps up the odds for creating rock steppes in otherwise dry areas to simulate fells/"badlands"/barren ecoregions, that used to be sand deserts instead.
3) Adds a latitude-dependent attenuation factor to the decrease of temperature with height in order to avoid the glacier spam that often happens.
    This attenuation factor might not be optimal though, but the equator regions will no longer feature enormous glaciers.

4) I might have done something very "unpythonic" or noobish by using "mcDup = ClimateMap()" to invoke the getLatitude function from the climate class. But I think it's okay?

### DIFF
--- a/PrivateMaps/FaireWeatherTweakEx.py
+++ b/PrivateMaps/FaireWeatherTweakEx.py
@@ -4976,7 +4976,7 @@ def generateShrubland():
             plot = mmap.plot(x,y)
             if not plot.getPlotType() == PlotTypes.PLOT_PEAK: #and (plot.isRiver() and not isAnyAdjacentPlotTerrainType(x, y, terrainDesert)):
                 if plot.getTerrainType() == terrainPrairie or plot.getTerrainType() == terrainPlains or plot.getTerrainType() == terrainGrassland:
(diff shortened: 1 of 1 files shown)


johan-buret (changes requested):
As a whole, the map doesn't generate enough desert tiles anymore, and by consequences it doesn't generate enough shrublands. 

Both tiles types are necessary from a game balance perspective.

The changes you propose are interesting though, and I feel like you are on the right track.

maushacks:
Shrublands should now spawn independently of deserts, they can still be quite scarce though, but colonising the New World was never meant to be easy after all.

Upping self.DesertPercent = 0.1 did pretty much alleviate the issue, but I reckon salt is meant to be retrievable even on maps that are entirely generated far north/south of the horse latitudes? I reckon this would either require to add the salt resource to the mountains or an entirely new landscape which raystuttgart once alluded to, to get it from the coasts.  I think the pull request should rather be a draft.

maushacks:
Requires a lot more thought on how to tackle the rest of the map terrains to make them both aesthetically pleasing and to not disturb gameplay.
I'll be back!

PR #944, opened 2023-11-10
Decrease Lush Grass feature "see through" parameter

victorxup:
It apparently shouldn't prevent visibility of adjacent tiles


raystuttgart:
Yeah, looks like a reasonable change. :)
Did you also check "meadows" or the cactus stuff, becauce it would make sense to be consisten then?

raystuttgart:
**Edit:**
Ok, I adjusted also the other Terrain Features like Meadows, Flowers, Shrubs, Winds .... that should not change see through.
(Thickets, Forrests, Storms, ... still do of course. Also Tornados and Volcanos now do.)

Issue #949, opened 2023-11-14
Poor AI initial ocean exploration


victorxup:
Recently I noticed that AI often delays founding its first city.
One of the reasons was that initial AI ship simply lost itself in the ocean and was busy with pointless exploration in the corners of the map instead of aimnig to the continents to settle down.
It appears that AI algorithm seems to be very sensitive to goodies and storm features. Once it deviates from its initial axis, things become very unstable.

This is not at all critical on small maps with small Distance to Europe.
But if you play Gigantic size with exteme Distance to Europe (>=8), then flaws of current AI become very clear: some of them fail to settle for many dozens of turns.

I already created a patch for this issue, and will make a pull request in a couple days, when it's properly tested.

PR #950, opened 2023-11-15
#949 Implement preferred starting direction for AI initial exploration

victorxup:
Improve AI starting exploration during first 10 turns (that number was hardcoded before, I didn't change this):
When on mission "explore ocean" starting ship now tends to aim itself to the direction, which is calculated basing on the surrounding of starting plot (how Europe tiles are situated, which Europe type they are).
The system is based on already implemented weight calculation system and should not break it.
The vessel still will choose next move basing not only on how close it is to new parameter "preferred direction", but also how many plots will be revealed and a bit of random.

**The result is:** 
On maps with big distance to Europe AI doesn't get lost in the ocean so often ;
-> reaches coast faster
-> first settlement is founded more quickly;
-> first settlement is in better location regarding distance to Europe

The implemented code can be improved a bit in part of performance, but it will increase complexity as well.
Also this code is executed only for the first several turns, so I think there is no need to bother.

**Remaining possible improvements (TODO):**
- teach AI to value tiles with unfavourable wind less than other tiles (it already succesfully avoids storm)
- maybe scale first 10 turns period depending on map size or game speed

### DIFF
--- a/Project Files/DLLSources/CvGameCoreUtils.cpp
+++ b/Project Files/DLLSources/CvGameCoreUtils.cpp
@@ -168,6 +168,93 @@ float directionAngle( DirectionTypes eDirection )
 	}
 }
 
+// returns minimal number of steps (clockwise or counterclockwise) from one direction to another
+int getDirectionDiff(DirectionTypes direction1, DirectionTypes direction2)
+{
+	FAssertMsg(direction1 >= FIRST_DIRECTION && direction1 < NUM_DIRECTION_TYPES 
+		&& direction2 >= FIRST_DIRECTION && direction2 < NUM_DIRECTION_TYPES,
+		"Invalid DirectionTypes enum argument");
+
+	// in case one of directions is NO_DIRECTION
+	if (direction1 == NO_DIRECTION || direction2 == NO_DIRECTION)
+	{
+		if (direction1 == NO_DIRECTION && direction2 == NO_DIRECTION)
+		{
+			return 0;
+		}
+		else
+		{
+			return 1;
+		}
+	}
+
+	int diff = std::abs(direction1 - direction2);
+	if (diff <= NUM_DIRECTION_TYPES / 2)
+	{
+		return diff;
+	}
+	else
+	{
+		return NUM_DIRECTION_TYPES - diff;
+	}
+}
+
+DirectionTypes getDirectionFrom_dX_dY(int dX, int dY)
+{
+	int x, y;
+	if (dX < 0) x = -1;
+	else if (dX > 0) x = 1;
+	else x = 0;
+
+	if (dY < 0) y = -1;
+	else if (dY > 0) y = 1;
+	else y = 0;
+
+	switch (x)
+	{
+	case -1:
+		switch (y)
+		{
+		case -1:
+			return DIRECTION_SOUTHWEST;
+		case 0:
+			return DIRECTION_WEST;
+		case 1:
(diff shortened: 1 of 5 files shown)


Nightinggale:
Please work based on the develop branch rather than master. We have a protected master, which isn't used for development, but rather will stay as the newest stable release version.

johan-buret (dismissed):
We can't merge like this what amounts to a full rewrite of CvUnitAI.cpp and CvPlayer.cpp

victorxup:
> We can't merge like this what amounts to a full rewrite of CvUnitAI.cpp and CvPlayer.cpp

Honestly, I didn't get what you mean.
Could you please get more into details? I'm open to suggestions.

I consider this change to be quite minor, but useful.

Voker57:
> > We can't merge like this what amounts to a full rewrite of CvUnitAI.cpp and CvPlayer.cpp
> 
> Honestly, I didn't get what you mean. Could you please get more into details? I'm open to suggestions.
> 
> I consider this change to be quite minor, but useful.

For some reason, your commit removes every line in these files and then adds them again. Not sure why this happened, maybe mass whitespace changes, but it should be fixed.

MrZorG33:
>maybe mass whitespace changes, but it should be fixed.

Indeed, in xml files they began to use tabs instead of whitespace. could this have an effect?

victorxup:
> For some reason, your commit removes every line in these files and then adds them again. Not sure why this happened, maybe mass whitespace changes, but it should be fixed.

hmm, OK, I'll look into this.
It's weird that github diff doesn't show that...

Voker57:
> > For some reason, your commit removes every line in these files and then adds them again. Not sure why this happened, maybe mass whitespace changes, but it should be fixed.
> 
> hmm, OK, I'll look into this. It's weird that github diff doesn't show that...

it does
diff

johan-buret:
@MrZorG33 the changes were brought to XML files, which are not involved in any way in this pull request.

@victorxup Github flags every single line of CvUnitAI.cpp and CvPlayer.cpp as changed, for no reason at all. Reviewing the changes in local show that, indeed, the effective change are much more reasonable in scope, like you told us.

victorxup:
Any idea what could be the cause? Maybe git cherry-picking? 
Okay, I'll handle this, force-push and notify you

Nightinggale:
Is it possible that there is an issue with line endings as in LF vs CRLF? I did set line endings .gitattributes, but it doesn't look like it is working as intended.

victorxup:
> Github flags every single line of CvUnitAI.cpp and CvPlayer.cpp as changed, for no reason at all. Reviewing the changes in local show that, indeed, the effective change are much more reasonable in scope, like you told us.

Fixed. Force pushed correct diff.
Pull Request can be reviewed now

johan-buret:
It looks much better this way.

johan-buret (approved):
The first tests are pretty conclusive. Even with a large coastal distance, the AI European first ships do not get lost

victorxup:
> The first tests are pretty conclusive. Even with a large coastal distance, the AI European first ships do not get lost

I usually play with 12 tiles distance, and AI is ok with it.
The tricky part is the rare cases and how bug-proof is this change.
Maybe we can add more asserts? But which?

johan-buret:
@Nightinggale  or @devolution79  are more at ease with the movement system than I am.

victorxup:
Just a kind reminder:
I suppose, this update can be merged to dev-branch and properly tested by community.

Issue #953, opened 2023-11-20
Naval unit heals negative promotion in open sea


victorxup:
To reproduce issue, just load save and skip turn

[AutoSave_AD-1492-May.zip](https://github.com/We-the-People-civ4col-mod/Mod/files/13419670/AutoSave_AD-1492-May.zip)

victorxup:
I suppose to fix this issue, the following section must do the **opposite**: 
AI heals neg.promotions everywhere, Human Player can heal neg.promotions only in settlements

![image](https://github.com/We-the-People-civ4col-mod/Mod/assets/23401480/46aedf7c-f265-4380-a6cf-9bdf1bf876e0)

johan-buret:
I confirm your analysis. I will patch that shortly

victorxup:
Or just merge my PR

PR #954, opened 2023-11-20
#953 Fix Human Player units healing negative promotions anywhere

victorxup:
Now AI can heal them everywhere, Human Player - only in settlements


MrZorG33:
Initially, "negative promotions" could be cured in the forts, colonies and villages of the natives. Will it still work the same now?

victorxup:
According to current implementation It will heal in **friendly or neutral** **settlements**, **forts**, **large forts** or **canals**

PR #972, opened 2024-01-09
Fix for second choice in "Pirate Attack" events

victorxup:
Now if player refuses to pay Pirate Ships are created


raystuttgart:
No idea what happened and why that had been commented out.

PR #978, opened 2024-01-14
New FF Competition Logic

Razonatair:
New logic for how the FFs are offered. Now it's a competition between all colonies that can hire that FF this turn. Whoever has more of the primary point type of that FF wins. If it's deadlocked, then we check the political points, and if those are deadlocked, then it goes to the first team called, but that will be extremely unlikely.

This code addresses the "France gets first dibs" problem, and also makes it so you can get an infinite number of offered FFs as long as you have the points available still. I've tested it already, and it works.


raystuttgart:
Ok, I merged this now.

PR #1060, opened 2025-01-01
Use the best of Europe and Africa prices while selling using Customs House

Voker57:
…House

### DIFF
--- a/Project Files/DLLSources/CvCity.cpp
+++ b/Project Files/DLLSources/CvCity.cpp
@@ -7411,16 +7411,19 @@ void CvCity::doYields()
 					int iLoss = std::max(GC.getCITY_YIELD_DECAY_PERCENT() * iExcess / 100, GC.getMIN_CITY_YIELD_DECAY());
 					iLoss = std::min(iLoss, iExcess);
 					changeYieldStored(eYield, -iLoss);
+					
+					// compare possible profits from selling Europe and Africa, and use the best possible
+					const int iEuropeProfit = GET_PLAYER(getOwnerINLINE()).getSellToEuropeProfit(eYield, iLoss);
+					const int iAfricaProfit = GET_PLAYER(getOwnerINLINE()).getSellToAfricaProfit(eYield, iLoss);
+					int iComparableProfitInEast = std::max(iEuropeProfit, iAfricaProfit);
 
 					// R&R, ray , Changes to Custom House - START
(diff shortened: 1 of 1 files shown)


Voker57:
One thing I'm concerned with this, is whether goods sold via CH will now count for both Africa and Europe. I was not able to find where is the code which records that.

raystuttgart:
Goods are counted towards **Europe** Sales once this is called:
`changeYieldTradedTotal(eYield, iAmount, iBuyValue)`

Goods are counted towards **Africa** Sales once this is called:
`changeYieldTradedTotalAfrica(eYield, iAmount, iBuyValue)`

Goods are counted towards **Port Royal** Sales once this is called:
`changeYieldTradedTotalPortRoyal(eYield, iAmount, iBuyValue)`

---

Meaning each Harbour (Europe, Africa, Port Royal) has its own function to be called to change the YieldTraded values.

Goods amounts trade influence 3 things:
* Taxes Algorithm
* Pricing Algorithm
* Trade Quests

---

**P.S.:**
There is 2 types of functions:
"Traded" and "Bought" ... but events and taxes all work with "Traded" if I remember correctly.
Pricing algorithm uses both to know if prices should increase or decrease.

But it has been a while since I worked with that code.

Voker57:
Thanks. I've modified the code to properly update the ones that were actually used in the trade

Nightinggale:
```cpp
if (iEuropeProfit > iAfricaProfit) {
} else {
}
```
Our coding style is to have brackets on their own lines, like
```cpp
if (iEuropeProfit > iAfricaProfit)
{
}
else
{
}
```
Makes no difference to the compiled dll, but it makes the code more readable if we stick to the already used standard.

Voker57:
Fixed. Is there an astyle config or style description somewhere?

Nightinggale:
> Fixed. Is there an astyle config or style description somewhere?

No. We are in dire need of somebody, who wants to organize the wiki and fill in all the stuff, which should be there. The main problem is that each hour spend on documentation is an hour not spend on the code and we don't have enough hours on the code as it is.

Voker57:
> > Fixed. Is there an astyle config or style description somewhere?
> 
> No. We are in dire need of somebody, who wants to organize the wiki and fill in all the stuff, which should be there. The main problem is that each hour spend on documentation is an hour not spend on the code and we don't have enough hours on the code as it is.

Well, if you at least verbally describe the style being used I can set up automated control of it

devolution79:
Just chiming in to add support for the "compact" style so I think that @Voker57's code is totally fine.

PR #1071, opened 2025-01-24
Automatic builds using Github Actions

Voker57:
This will build a mod archive from every commit to specified branch(es), and make a pre-release with file attached to it. You can see how it works here https://github.com/Voker57/wtp/releases/tag/development

Potential improvements:

* [ ] A release is a >1GB file with most data unchanged. Nightingale suggested using some kind of incremental releases, this can be implemented if there's an agreement on how it should be done
* [ ] Fastdep is not used because it doesn't work for some reason. It could theoretically speed up builds.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,45 @@
+on:
+  push:
+    branches:
+    - develop
+jobs:
+  my_job:
+    runs-on: windows-2025
+    name: wtp
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          path: build
+      - name: getcompiler
+        uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


Voker57:
I have replaced the script hacks with use of environment variables to avoid asking for input. Incremental releases, after consideration, would require a rather complicated setup with questionable value, so this will be excluded from this PR. The request is ready to merge.

Nightinggale:
I'm not sure using compile.pl is the right move for this, but given that it reverts back to normal without specific environmental variables, I suppose it's doable. I will take a closer look at this when I get time to do so.

Voker57:
Superceded by https://github.com/We-the-People-civ4col-mod/Mod/pull/1122

PR #1077, opened 2025-02-03
Founding Fathers Name Updates

mistakingover:
Fixing in different language names for Founding Fathers.
Johann Moritz line (4108) left intact, but his name in code should be changed to reflect the correct english name, akin to the standirization used with other fathers.
As I have no knowledge on how this would affect other areas of the code, left it untouched, so only things that changed were how the names show in-game.
Johann Moritz Pedia text also changed to reflect his Dutch name in all entries, as common in other Founding Fathers, having their original language name in the Pedia description.
Also notice that the title Furst (Prince in German) was also changed in all lines to reflect the correct language of the text it's writen on, ending all possible mistakes when reading it.
-----
More name language updates. Stopped since there seems to be a need in better representing the various nations of the continent, which deems updating FF's names to their correct names in different languages potentially redundant given the possibility of removal.
Will address such matters when the list of FF's is definite.

### DIFF
--- a/Assets/XML/Text/XML_AUTO_UTF8_FatherInfo.xml
+++ b/Assets/XML/Text/XML_AUTO_UTF8_FatherInfo.xml
@@ -4067,7 +4067,7 @@
 	<TEXT>
 		<Tag>TXT_KEY_FATHER_WILLEM_BARENTS</Tag>
 		<English>
-			<Text>Willem Barentsz</Text>
+			<Text>William Barentsz</Text>
 			<Gender>Male</Gender>
 			<Plural>0</Plural>
 		</English>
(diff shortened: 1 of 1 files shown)


MrZorG33:
undo the change for the TXT_KEY_FATHER_JOHANN_MORITZ tag for Russian and for the TXT_KEY_FATHER_IVAN_KRUZENSHTERN tag for all languages.
1. I am translating into Russian based on the information I have (in Russian), and if something needs to be corrected, I will do it myself.
2. Although Kruzenshtern had the name you indicated at birth, he took a Russian name in his youth and was subsequently called by that name everywhere, since he was a Russian navigator, although he was not one at birth. His original name can be indicated in the description.

Issue #1116, opened 2025-04-12
Emphasize / Deemphasize seems not to be working. (Candidate pull request seems to "fix" the issue).


abmpicoli:
**Describe the bug**
It seems that the Emphasize/Deemphasize buttons have no effect in automatic allocation.

**To Reproduce**
In the save provided, enter the Unalashka town. I see the free colonist working on geting furs. 4 furs + 2 food.

![Image](https://github.com/user-attachments/assets/5025948d-5a96-4349-8c16-2c5a8656d468)

Now, I deemphasize food and furs, and emphasize, say, sheep:

![Image](https://github.com/user-attachments/assets/e2ccd174-2770-41d1-ae8c-b0339cb9a6ff)

![Image](https://github.com/user-attachments/assets/5a22561a-998c-4b84-9922-ac70c90d1b03)

![Image](https://github.com/user-attachments/assets/94693c7d-3d41-4d91-88bb-3d5691b617c4)

Click the "Turn on city automation button".

Ideally provide a savegame where this happens consistently, in which case the steps to reproduce would be load, click end turn or something like that.

**Expected behavior**
Some of the colonists working on fur should work on sheep.
The behavior is the same is I emphasize other things, such as hammers, or stone: there are no carpenters working or people herding sheep

![Image](https://github.com/user-attachments/assets/5ea8de77-bcac-4874-a271-5d9684e39155)

![Image](https://github.com/user-attachments/assets/f912bcfc-406e-4ad2-9aed-05e33261a059)

![Image](https://github.com/user-attachments/assets/72c03606-8925-4e3d-a25d-3a856c732630)

[Ale Xan De AD-1570-Emphasize-Deemphasize not working.zip](https://github.com/user-attachments/files/19719451/Ale.Xan.De.AD-1570-Emphasize-Deemphasize.not.working.zip)

**Personal modifications/We The People version**

```
built from source: changes to the text localization file and creation of a new map.

$ git status
On branch abmpicoli_changes
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   Assets/XML/Text/CIV4GameText_TAC_utf8.xml

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        PrivateMaps/FaireWeatherTweakExAbmpicoli.py

no changes added to commit (use "git add" and/or "git commit -a")

[...]

PR #1117, opened 2025-04-12
Possible fix for issue #1116

abmpicoli:
It seems that the reason why emphasize yields doesn't work is because they have been isolated from code.

This fix makes the enphasize yields to be added as a final calculation to yield value.

I've also changed the code that makes Cannon values to be set to 0 to be only if they are either Normal, or de-emphasized.


abmpicoli:
Update: it seems that there is some extra thing happening with the emphasize/deemphasize logic: So my own recommendation is to hold this pull request for a moment. 

I will explain details in the related issue.

UPDATE: it seems it is not related to this logic, but a bug in how professions are valued: only the primary yield is considered for assigning profession value, so hunters aren't considered as food sources. Opened https://github.com/We-the-People-civ4col-mod/Mod/issues/1118 for that

abmpicoli:
Just confirming, the change is working locally, and the issue I've found is not related to the change, but to existing issue #1118. So I consider this pull request ready for review.

abmpicoli:
Closing the pull request since it has some personal hacks accidentally pushed into it.

PR #1122, opened 2025-04-30
Add build + pre-release Github workflow mk2

Voker57:
This will build a mod archive from every commit to specified branch(es), and make a pre-release with file attached to it. You can see how it works here https://github.com/Voker57/wtp/releases/tag/development. This is a new PR to change the poorly chosen source branch.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,45 @@
+on:
+  push:
+    branches:
+    - develop
+jobs:
+  my_job:
+    runs-on: windows-2025
+    name: wtp
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          path: build
+      - name: getcompiler
+        uses: actions/checkout@v4
+        with:
+          path: Compiler
+          repository: We-the-People-civ4col-mod/Compiler
+      - name: cache
+        uses: actions/cache@v3
+        with:
+          key: wtpcache
+          path: |
+            perl
+      - name: build
+        run: |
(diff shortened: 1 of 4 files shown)


raphaelmonticello:
Hey @Voker57

I looked at this because I heard you are unhappy it sat here so long. That is fair. From my side the problem is not that nobody cares. It is that most people here do not understand what this PR is even about.

The title talks about a "Github workflow". The description is two sentences plus a link to your own repo. For people who review ships, events and city screens, that reads like a black box. They skip it because they do not know what they would be merging.

If you want this to move, please rewrite the first post so a player or a regular WTP modder can follow it. No CI jargon. Something like:

1. What problem does this solve?
   Example: testers can download a ready-to-play zip of current `develop` without compiling the DLL on their PC.

2. What happens after it is merged?
   - every push to `develop` builds a zip
   - it updates the GitHub pre-release called "development"
   - testers download it from Releases, unzip into `Mods`, start the game
   - this is not an official release, just a snapshot of the branch

3. How do I try it today, step by step?
   Link the demo zip. Then: download where, put the folder where, how to launch. Assume the reader never opened GitHub Actions.

4. What does this change in our compile scripts?
   Say clearly that local compile stays the same, and the new env vars only kick in on the server.
   Also say which DLL type CI builds (right now `compile.pl` is forced to option 6, Hardcoded Final Release) and why that one.

5. What this does NOT do
   Not a gameplay change. Not a smaller incremental zip. Nightinggale already asked if `compile.pl` is the right tool — answer that in the description so people do not have to dig #1071.

6. Change the title to human words

PR #1137, opened 2025-06-07
possible fix for issue #1116

abmpicoli:
Fix for issue #1116, where emphasize/deemphasize doesn't work


Schmiddie1979:
I revisited this issue and was able to reproduce the reported behavior on the current codebase.

Emphasize/deemphasize was not completely without effect, but automated citizen assignment often reacted in ways that did not properly reflect the selected priorities.

Based on the general approach proposed in #1137, I implemented and tested an updated fix against the current codebase. The emphasize/deemphasize settings are now taken directly into account when evaluating yields for automated citizen assignment.

In my tests, this makes the automated profession assignment much more consistent and predictable, and the issue appears to be fixed.

I intentionally left out the additional YIELD_CANNONS change from #1137 for now, as the current AI handling of cannons deserves a separate investigation.

Issue #1124, opened 2025-05-09
Docks for fresh water lake/river towns.


RussMaster2299:
Fresh Water lakes and rivers usually have more food value to them compared to coastal and ocean tiles. While that is still true on a base level, being unable to build a Pier or Dock in a town that doesn't have ocean access limits the these fresh water lakes and rivers if you don't build next to the ocean. 

Please let us build up to a dock in these freshwater towns, so that there's an incentive to settling near a large river, or freshwater lake.
You could also allow a fish farm improvement to be built on fresh water rivers, or inland lakes, but with the lake requiring a boat to enter anyway, it makes more sense to just settle with the dock.

If you can upgrade it to a Drydock, This would also allow you to build fishing ships for large freshwater lakes, so building on the edge of these lakes can yield more food.

Previously, I made a post asking for the feature, but at the time, I wasn't aware that freshwater lakes and rivers didn't allow you to build the Pier, even if you were built next to them. I felt the need to clarify here.

abmpicoli:
I've just found this issue myself. It is a bug: according to the file CIV4BuildingInfos.xml, piers and docks should be allowed near lakes/rivers/ice lakes: however in the entry there is a requisite "`<bWater>1</bWater>`" . This requisite seems to be only valid for shallow coasts/coasts. Hence it is impossible to build them at those cities. At first glance, the solution is simple: Make the Civ4BuildingInfos.xml to have <bWater>0</bWater> for both the pier and dock buildings. Attaching the file where I did the fix for my local build

[CIV4BuildingInfos.zip](https://github.com/user-attachments/files/20640011/CIV4BuildingInfos.zip)

Just replace the file assets/xml/buildings inside the mod with the one in the zip file.

The diff for that file is

```
diff --git a/Assets/XML/Buildings/CIV4BuildingInfos.xml b/Assets/XML/Buildings/CIV4BuildingInfos.xml
index d1a42f15..778cf1ee 100644
--- a/Assets/XML/Buildings/CIV4BuildingInfos.xml
+++ b/Assets/XML/Buildings/CIV4BuildingInfos.xml
@@ -3128,7 +3128,7 @@
                        <FreePromotion>NONE</FreePromotion>
                        <bGraphicalOnly>0</bGraphicalOnly>
                        <bWorksWater>0</bWorksWater>
-                       <bWater>1</bWater>
+                       <bWater>0</bWater>^M
                        <bRiver>0</bRiver>
[...]

PR #1138, opened 2025-06-07
possible fix for issue #1124

abmpicoli:
The Civ4BuildingInfos.xml clearly states that the pier/dock can be built next to lakes, but the flag <bWater>1</bWater> forbids building around river/lakes.

At first view, the simplest solution is to say that piers/docks has bWater set to 0.


Schmiddie1979:
No longer required.

PR #1208, opened 2026-03-07
In a domestic advisor screen, displays building or unit in red if there are not enough resources to finish the building or unit

Ellestar:
In a domestic advisor screen, displays building or unit in red if there are not enough resources to finish the building or unit

### DIFF
--- a/Assets/Python/Screens/CvDomesticAdvisor.py
+++ b/Assets/Python/Screens/CvDomesticAdvisor.py
@@ -411,11 +411,11 @@ class CvDomesticAdvisor:
 							szText = "<color=255,0,0>" + szText + "</color>"
 							break
 			elif (pLoopCity.isProductionUnit()):
-				building = pLoopCity.getProductionUnit()
+				unit = pLoopCity.getProductionUnit()
 				pPlayer = gc.getPlayer(CyGame().getActivePlayer())
 				for iYield in range(YieldTypes.NUM_YIELD_TYPES):
 					if gc.getYieldInfo(iYield).isCargo():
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
I like the general idea of this PR, as it makes it much easier to identify cities where production is missing required goods.

However, I think there is one important case which might be worth considering before merging this.

As far as I understand the current implementation, it only checks the currently active building or unit. In WTP, production already invested into a building or unit is retained when switching to another production item. This means that a project may already have all (or most) of its required hammers, but cannot be completed because some required goods are missing. The player can then switch to another project and return to the unfinished one later once the missing goods are available.

These inactive but already progressed projects are actually among the cases where such an indication would be particularly useful. At the moment, however, they would no longer be visible because getProductionBuilding() / getProductionUnit() only refer to the currently active project.

Perhaps the feature could therefore be extended to also consider buildings and units with stored production progress and indicate when they are waiting for required goods. That would make the Domestic Advisor much more useful for finding projects which could be completed once the missing resources are supplied.

The original idea is definitely useful, so if the original author is no longer active, perhaps somebody else would like to pick this up and extend it.

raphaelmonticello:
Picked this up with Schmiddie's stored-production case added.

Ellestar's original branch is on `Ellestar:develop`, so this is a new PR from current `develop` instead of a push onto this one:
https://github.com/We-the-People-civ4col-mod/Mod/pull/1280

What #1280 does:
1. current production still goes red when cargo is missing (Ellestar's idea).
2. parked buildings/units with stored hammers that still need goods are listed in red brackets after the current item, e.g. `Caravel (4) [Drydock, Wagon]`.

This PR can be closed once #1280 is merged.

Issue #1205, opened 2026-02-26
CvRandomInterfaceEvent triggers a tuple out of range exception


antollare:
**Describe the bug**
error popup:

<img width="1250" height="765" alt="Image" src="https://github.com/user-attachments/assets/e32d8c62-9d91-4e43-9711-3c699784a192" />

Python Exception
Traceback (most recent call last):

File "CvRandomEventInterface", line 1084, in
canTriggerVolcanoFormant1

IndexError: tuple index out of range

In most recent case this pop up appears at the start of the turn after turn 225 on marathon speed in a generated map. It has occurred on the americas gigantic map, and the west indies huge maps. The pop up appears on every, or close to every turn after occurring once.

The popup can be closed and there doesn't seem to be any adverse affects. This error continues to occur after loading the save

**To Reproduce**
To get error from new game:
1. Create map and play game in marathon speed to around turn 250.
2. Error should occur at some point after ending turn

To get from attached save file:
1. Load "error save" game
2. End the turn
3. Error should occur

**Expected behavior**

**Screenshots and savegames**

<img width="1250" height="765" alt="Image" src="https://github.com/user-attachments/assets/9cc4744a-e215-4526-ab5d-53d4e2f60cb1" />

savegame:

[tupleError.zip](https://github.com/user-attachments/files/25588470/tupleError.zip)

**Personal modifications**
Nothing changed

**Version of We The People**
WTP 4.2.1

CirqueDuMal:
I am getting the same error, same version, also playing on Marathon

Nightinggale:
I have located the bug and is in progress of reviewing the code to ensure it is fixed properly this time. We had this before and apparently the fix wasn't good enough.

Sassazin:
I also have this issue as described, and it renders the game unplayable since it happens every time on Amrayca map, after 200+ turns.

~~Trying to apply the fix myself led to "Failed to initialize python"~~
(Windows 11 default permission issues on My Documents...)

Applying the fix on from https://github.com/acato/Mod/commit/504835b311aedc40a8f6e24c0a8a061df9d7166e on an existing save the turn right before the bug occurs does **not** fix it. 

Steps:
1) perform clean reinstall of Civ4Col including deleting registry entries + AppData + MyDocuments files
2) get the latest version from git repo and place in [Gamepath]\Civ4Colonization\MODS\WeThePeople
[...]

PR #1217, opened 2026-03-27
Fix event random numbers never stored in setRandomNumbers()

acato:
## Summary

- `EventTriggeredData::setRandomNumbers()` creates a `RandomContainer` for each event, generates a random number, but never calls `m_RandomNumbers.push_back(container)` — the vector stays empty
- This causes `getRandomNumber()` and `getRandomNumberForIndex()` to always return 0 (the fallback default)
- All probability checks using these functions become no-ops (e.g., `canTriggerVolcanoDormant1` fires 100% instead of ~25%)

## Details

The bug affects two code paths:

1. **Python callbacks** using `getRandomNumberForIndex()` — e.g., the volcano dormant event always triggers because `0 < 250` is always true
2. **DLL-side `TriggerChance` fastpath** in `CvPlayer::canDoEvent()` which uses `getRandomNumber(eEvent)` — always returns 0, bypassing probability gates

The fix is a single missing `push_back` call.

## Test plan

- [ ] Start a game with a volcano on the map; verify the dormant event doesn't fire every single time
- [ ] Verify events using `TriggerChance` fire at their intended probability rather than always/never

Related: #1205

### DIFF
--- a/Project Files/DLLSources/Events/EventTrigger.cpp
+++ b/Project Files/DLLSources/Events/EventTrigger.cpp
@@ -71,6 +71,7 @@ void EventTriggeredData::setRandomNumbers()
 			RandomContainer container;
 			container.event = eEvent;
 			container.number = GC.getGameINLINE().getSorenRandNum(1000, "Event random number");
+			m_RandomNumbers.push_back(container);
 		}
 	}
 }


devolution79:
@Nightinggale could  you have a look at this ?

Schmiddie1979:
I assume we should do this fix. It could have influence on the existing events, but I have checked if there are any events which use probabilities via 

1. TriggerChance(argsList)
2. canTriggerVolcanoDormant1
3. getRandomNumber()

but I could not find any or only the volcano event and very few that still use TriggerChance. Hence the effect should be limited. We need to monitor if any events have unintended issues after the fix.

Schmiddie1979:
content of the commit already exists in the current develop branch

PR #1220, opened 2026-03-29
Fix setProfession assert failures in AI city management

acato:
## Summary
- **AI_swapUnits infinite loop**: When `setProfession` failed during a unit swap, the code entered `while (!bResB)` retrying the same call forever (`bResB` was `const`). Replaced with `canHaveProfession` pre-checks and a fallback that restores original professions on failure.
- **setUnitWorkingPlot missing validation**: Only checked `isProfessionValid` (civilization-level) before assigning a plot profession, not `canHaveProfession` (which also checks equipment/slot requirements). Added `canHaveProfession` to the profession filter loop.

## Test plan
- [ ] Play several AI turns and confirm no setProfession asserts fire
- [ ] Verify AI units correctly swap professions in cities
- [ ] Verify units assigned to work plots have valid professions


Schmiddie1979:
Thanks for the PR. I agree with the underlying idea and think the additional validation is a sensible way to prevent invalid profession assignments and the potential loop in AI_swapUnits().

I have therefore implemented and tested the relevant changes as a separate fix in the current develop version.

I decided not to merge this PR directly, though, because it contains extensive changes to several files beyond the actual fix. Since some of those files have changed considerably in the meantime, merging them as a whole could introduce unrelated or difficult-to-predict side effects.

I therefore kept the current code wherever possible and incorporated only the changes needed to address the issue.

Thanks for identifying the problem and providing the proposed solution — it was very helpful in getting this fixed.

PR #1221, opened 2026-03-29
Fix crash when natives send ships to Europe (#1039)

### LINKED ISSUE #1039: Natives crash the game by sending ships to Europe if they get a ship, which is able to leave the map
I don't know what is causing, I tried a few things to fix it. Nothing works as of yet. All I know is that going back to a further back save and doing differing things doesn't fix it.

This a link to the exact turn that would crash after you end turn. I haven't had this happen I'd prefer not have to restart this run.
https://drive.google.com/file/d/1wJE7il63kuTbqmLDdhrA5L2HkuT8mdSN/view?usp=drive_link

acato:
## Summary
- Natives can acquire ocean-going ships (e.g. via gifting a Slave Barque). The ship's AI then sails to Europe, but natives have no Europe dock — the game crashes when accessing non-existent dock units.
- Added `isNative()` early-return checks to `canCrossOcean()`, `canSailToAfrica()`, and `canSailToPortRoyal()` so native-owned ships are blocked from leaving the map entirely.

Fixes #1039

## Test plan
- [ ] Gift an ocean-capable ship to a native tribe via World Builder
- [ ] Advance turns and confirm no crash (ship should remain on the map)
- [ ] Verify European players can still sail to Europe/Africa/Port Royal normally


Project Files/DLLSources/CvUnit.cpp
```diff
@@ -5024,6 +5024,12 @@ bool CvUnit::canAutoCrossOcean(const CvPlot* pPlot) const
 
 bool CvUnit::canCrossOcean(const CvPlot* pPlot, UnitTravelStates eNewState) const
 {
+	// Natives have no Europe/Africa/Port Royal — prevent crash when they acquire ocean ships
+	if (isNative())
```
devolution79: I would greatly prefer checking whether the player either 1) has a parent, and is therefore a European player, or 2) is a king. See CvPlayer::isColonialNation() and the related identity functions.

Kings do not really interact with Europe in the normal sense. Their ships are spawned there after independence is declared, but the king does not currently trade with Europe after that. Even so, treating the king as conceptually connected to Europe is still the correct abstraction, especially if we later implement this feature for kings as well.

Another advantage of not checking directly whether the player is native is that this also covers other player types, such as wild animals, i.e. the barbarian black team.

I am not going to add the same comment to the other two functions, but they should obviously use the same check(s).

A separate question is how the native player got hold of the Slave Barque in the first place. That is worth investigating and locking down, since it makes little sense given the AI's capabilities and the general oddity of natives having an ocean-going vessel. In short I think it should be banned, but that is a topic for another ticket.

devolution79 (changes requested):

Project Files/DLLSources/CvUnit.cpp
acato: You are 200% right. I had not thought about barbarians and wild animals as special cases. I updated the PR with the check you suggested.

Schmiddie1979:
Thanks for reporting and investigating this issue. The underlying bug is valid, and the work done here was helpful in identifying the actual problem.

However, I have decided not to merge this PR in its current form.

The main reason is the very large scope of the PR compared to the original issue. It contains a considerable number of changes across different parts of the codebase, including changes which affect fairly fundamental game mechanics and are not directly required to fix #1039.

With changes of this size and scope, it becomes increasingly difficult to understand all interactions with existing WTP mechanics and to reliably assess possible side effects. This also makes reviewing and testing the actual bugfix much harder. For this kind of issue I would strongly prefer a small and targeted change which leaves unrelated existing logic untouched.

That said, the discussion here provided an important part of the solution.

In particular, I agree with the suggestion that this should not simply be handled as a special case for Native players. The actual question is which player types are conceptually allowed to use the off-map travel mechanics in the first place.

I therefore used the positive check suggested in the discussion:

```cpp
const CvPlayer& kOwner = GET_PLAYER(getOwnerINLINE());
if (kOwner.getParent() == NO_PLAYER && !kOwner.isEurope())
{
	return false;
}
```

This allows colonial players with a parent and Europe/King players, while automatically excluding Natives, Barbarians/Wild Animals and other player types which currently have no connection to the off-map markets.

During testing I found one additional issue which also needed to be taken into account. Blocking only the actual travel functions was not sufficient: an ineligible ship could still enter the "Sail to Europe/Africa/Port Royal" automation and could then become unable to move.

The same eligibility condition is therefore also checked when those automation modes are made available. As a result, an ineligible player cannot start the corresponding automation in the first place, and the buttons are no longer shown.

I also agree with the point about possible future features for Kings or other player types. The current check deliberately represents the game rules as they exist today; it should not be understood as a permanent definition of which player identities may ever use off-map trade.

If WTP later allows additional player types – for example Pirates or other special factions – to interact with Europe, Africa or Port Royal, I think the better long-term solution would be to introduce explicit player-level capability functions for this purpose rather than adding more identity checks throughout `CvUnit`, for example conceptually:

```cpp
canUseEurope()
canUseAfrica()
canUsePortRoyal()
```

The unit and automation code could then ask whether the player has the relevant capability, while the definition of that capability would be maintained in one place. This would also allow the three off-map markets to develop different access rules in the future.

I don't think introducing such a larger abstraction is necessary as part of this bugfix, however. For now I prefer to keep the change small and fix the concrete problem with the current game rules.

The separate question of how a Native player can obtain an ocean-going vessel such as the Slave Barque is also worth investigating, but I agree that this should be handled separately rather than mixed into this fix.

So I will close this PR, but thank you for finding the bug and working on a solution. The investigation and discussion here were useful and directly contributed to the smaller fix which I have now implemented and tested.

PR #1222, opened 2026-03-30
Fix crash from unit in city and on map simultaneously (#923)

### LINKED ISSUE #923: Removing and adding a citizen from a colony leaves the unit on the plot outside the colony
I might finally have figured out the cause of NULL units in selection groups.

When loading the attached savegame and ending the turn, unit 1 of player 37 (expert farmer in Fort Christina) will call CvPlot::addUnit twice, but never calls CvPlot::removeUnit. The game then crashes in CvUnit::setInfoBarDirty as the unit instance is the farmer, but since it's not on the map, it's NULL.

Why the farmer is added to the plot twice and why it's not removed when rejoining the city is still unknown.

[jhxp_AD-1517-January7.zip](https://github.com/We-the-People-civ4col-mod/Mod/files/13126904/jhxp_AD-1517-January7.zip)

acato:
## Summary
A unit in a city's population list can end up simultaneously present on the map plot (the exact root cause of the dual state is unknown — multiple devs investigated). When `removePopulationUnit` then calls `addToMap`, `CvPlot::addUnit` inserts the unit a second time. The game later crashes accessing the stale duplicate reference.

Three defensive guards prevent the crash:

- **`CvPlot::addUnit`**: While walking the unit list to find the insertion point, check if the unit is already present. If so, assert and return early — prevents the double-add that causes the crash.
- **`CvCity::addPopulationUnit`**: If `getAndRemoveUnit` returns NULL (unit not in the player's unit list, likely already in another city), bail out instead of pushing NULL into `m_aPopulationUnits`.
- **`CvCity::removePopulationUnit`**: Before calling `addToMap`, check if the unit already has valid plot coordinates. If it's already on the map, skip `addToMap` to prevent double-add.

These are defensive checks — the assert messages will fire in Assert builds to help track down the original state corruption if it still occurs.

Fixes #923

## Test plan
- [ ] Load the attached savegame from the issue and end turn — verify no crash
- [ ] Play several turns with AI cities managing workers — verify no asserts fire
- [ ] Verify human player can normally add/remove citizens from colonies


Schmiddie1979:
I took a closer look at #923 and at the existing population-unit transfer code before deciding how much of #1222 we actually need.

The underlying problem described in #923 is that a unit can apparently end up in an inconsistent state where it is still treated as a city population unit, but is already present on the map as well. The exact root cause of that state is still unclear.

There is already an older recovery mechanism in `CvCity::doTurn()` which was added by Nightinggale:

```cpp
// Fixing stuck units in colonies - Nightinggale - Start
// Once in a while a citizen can lose profession in a colony
// ...
// Doesn't fix the cause of the problem, but this way the player will regain lost units during new turn event
```

That code searches for population units with `NO_PROFESSION` and ejects them from the colony by calling `removePopulationUnit()`.

While looking at this, I noticed that `removePopulationUnit()` itself assumes that the population unit is always completely off-map before it is transferred back to the player:

```cpp
GET_PLAYER(getOwnerINLINE()).addExistingUnit(pUnit);
pUnit->addToMap(coord());
pUnit->setProfession(eProfession);
```

Normally that assumption is correct. A proper population unit has invalid map coordinates and therefore needs to be added back to the map when it leaves the colony.

However, if the inconsistent state from #923 has already occurred and the population unit is somehow still present on the map, the unconditional `addToMap()` tries to add the same unit to the plot a second time. That appears to be the immediate cause of the duplicate plot-unit state and the subsequent crash.

I therefore changed this to:

```cpp
GET_PLAYER(getOwnerINLINE()).addExistingUnit(pUnit);

if (pUnit->getX_INLINE() == INVALID_PLOT_COORD &&
	pUnit->getY_INLINE() == INVALID_PLOT_COORD)
{
	pUnit->addToMap(coord());
}

pUnit->setProfession(eProfession);
```

This keeps the normal behavior unchanged: a regular population unit is still added to the city plot when it leaves the colony.

If the unit is already on the map because of the bug, however, it is not added a second time. This also makes the existing Nightinggale recovery code safer, because it can now eject a broken/stuck citizen without potentially triggering the duplicate `addToMap()` problem itself.

I also added a second defensive check in `addPopulationUnit()`.

The current code does:

```cpp
CvUnit* pTransferUnit = GET_PLAYER(pUnit->getOwnerINLINE()).getAndRemoveUnit(pUnit->getID());
FAssert(pTransferUnit == pUnit);
```

and then immediately continues using `pTransferUnit`.

If `getAndRemoveUnit()` unexpectedly returns `NULL`, the assert catches this in a debug build, but the code would still continue afterwards and eventually dereference a null pointer.

I therefore added:

```cpp
if (pTransferUnit == NULL)
{
	return;
}
```

This does not change the normal transfer path, but prevents another crash if the unit is already in an invalid state when it is moved into the colony.

I deliberately did **not** add the additional global guard in `CvPlot::addUnit()` from #1222 for now. That would affect a much broader and more fundamental part of the unit system. Since the two `CvCity` changes directly protect the population-unit transition where the problem occurs, I prefer to keep the fix as small and local as possible.

I also tested the critical case manually. Since the original save from #923 is likely too old to load with the current development version, I temporarily modified `removePopulationUnit()` to simulate the broken state by putting the population unit on the map before the normal transfer-back code was reached.

With the old logic this would result in a second `addToMap()` call.

With the new check, the second call is skipped. The game continued normally and the unit remained fully usable on the map.

After removing the temporary test code, I also tested the normal case again by moving citizens out of a colony. That continues to work normally.

So the result is:

- normal population-unit transfers are unchanged;
- `addPopulationUnit()` no longer continues with a null transfer unit;
- `removePopulationUnit()` no longer adds an already mapped unit to the plot a second time;
- Nightinggale's existing stuck-citizen recovery code benefits from the same protection automatically;
- no changes were made to the broader `CvPlot::addUnit()` logic.

This does **not** identify the original root cause of why a citizen can occasionally end up both in the colony and on the map. That part of #923 remains unexplained.

However, the dangerous inconsistent state is now handled defensively at the city/population boundary, and the duplicate-map-unit crash path described in #923 should be prevented.

PR #1223, opened 2026-03-30
Exclude traveling units from combat and space calculations (#796)

### LINKED ISSUE #796: Ship arriving from Europe on next turn is involved in combat
The attached savegame has the following situation:
[33,71] has a ship A (dutch fluyt, "Delfland")
[33,71] also has a ship B that will arrive on the next turn from europe (english caravel, "Suffolk")

Attacking the plot with the players privateers somehow involves ship B in the combat.
There should be a check whether the ship is en route to/from Europe before involving it in combat.

The bug is probably very old, but was just never noticed ...
[save.zip](https://github.com/We-the-People-civ4col-mod/Mod/files/11177990/save.zip)

acato:
## Summary
Units traveling to/from Europe/Africa/Port Royal keep their plot coordinates (at the map edge) but should not interact with gameplay on that tile. A ship in transit could be "involved in combat" when enemies attacked the tile.

- **`canDefend()`**: Added `isOnMap()` check. While `getBestDefender()` already filtered via `isOnMap()`, the lower-level `canDefend()` did not — so `getNumDefenders()` (used by AI movement decisions) and any direct callers of `canDefend()` would incorrectly count traveling units as defenders.
- **Harbour/barracks space**: The `canMoveInto()` harbour and barracks space calculations looped through all plot units without checking travel state. Traveling ships/troops incorrectly consumed space, potentially blocking valid movement.

Fixes #796

## Test plan
- [ ] Build Assert DLL — no compilation errors
- [ ] Verify ships in transit to Europe are not counted as defenders (attack a tile with a traveling ship)
- [ ] Verify harbor space is not consumed by ships in transit
- [ ] Verify normal combat still works correctly


Project Files/DLLSources/CvUnit.cpp
```diff
@@ -9741,6 +9741,12 @@ bool CvUnit::canDefend(const CvPlot* pPlot) const
 		pPlot = plot();
 	}
 
+	// Units traveling to/from Europe/Africa/Port Royal cannot defend
+	if (!isOnMap())
```
devolution79: Thanks, this could very well be the fix for this issue. Please use the inline isOnMap_()  variant.

Project Files/DLLSources/CvUnit.cpp
```diff
@@ -3499,7 +3499,7 @@ bool CvUnit::canMoveInto(CvPlot const& kPlot, bool bAttack, bool bDeclareWar, bo
 						// we only count Land Units that can attack, civil Units are not considered
 						// we also not consider Units loaded on Ships
 						// we also not consider Units of other Nations
-						if (pLoopUnit != NULL && pLoopUnit->getDomainType() == DOMAIN_SEA)
+						if (pLoopUnit != NULL && pLoopUnit->getDomainType() == DOMAIN_SEA && pLoopUnit->isOnMap())
```
devolution79: These  additional checks do not belong here because they serves no purpose. A unit for which isOnMap() is false should never reach canMoveInto() unless there is already some serious bug elsewhere.

In fact, if such a check were needed at all, the correct place for it would be much higher up in the call chain. Units that are not on the map are excluded by the normal unit move action processing loop, except for the special case of units in port, where the port update function is called directly rather than going through CvUnit::AI_update().

Just revert the changes to this function.

devolution79 (changes requested):

acato:
Thanks - Updated the PR with both changes.

Schmiddie1979:
Thanks for identifying this issue and for the proposed fix.

I tested the underlying problem and confirmed that canDefend() is indeed called for units which are currently off-map/traveling. I therefore added the isOnMap_() check to CvUnit::canDefend() separately to develop.

I also tested the change in-game, including explicitly asserting calls to canDefend() for off-map units, and the check is triggered as expected. After removing the test assert, everything has been running without issues so far.

I don't want to merge this PR as a whole, however, since it contains a number of additional and unrelated changes which would make it difficult to assess possible side effects.

So the actual fix for #796 has been incorporated separately.

PR #1224, opened 2026-03-30
Fix World Builder gold editing applied to wrong player (#1045)

### LINKED ISSUE #1045: WB: gold is given to the wrong player
Editing a city will set gold for the active player rather than the city owner.
Assets/Python/Screens/CvWorldBuilderScreen.py:1336 shows part of the problem, but it is indeed just part of it.

acato:
## Summary
- The city editor's gold spinner used `m_iCurrentPlayer` (the active/selected player) for both reading the current gold value and writing changes. When editing a foreign city, gold was displayed and modified for the wrong player.
- Changed both the spinner initialization (line 1345) and the callback handler (`handleCityEditGoldCB`) to use `self.m_pActivePlot.getPlotCity().getOwner()` — the actual city owner.

Fixes #1045

## Test plan
- [ ] Open World Builder, select a foreign city, verify gold spinner shows that player's gold
- [ ] Change gold value, verify it applies to the city owner (not the active player)
- [ ] Verify editing your own city's gold still works correctly


Schmiddie1979:
This is not a bug. You can change the player in the WB you want to assign gold respectively you want to give gold. 

Change not necessary.

PR #1227, opened 2026-04-03
Fix remote crash from negative PlayerTypes in CvNetwork::receive()

### LINKED ISSUE #1225: Security: Negative PlayerTypes in CvNetwork::receive() causes remote crash in multiplayer
## Bug Description

In `CvNetwork::receive()` (`Project Files/DLLSources/Network/CvNetwork.cpp:56-80`), the `ePlayer` parameter is a `PlayerTypes` (signed enum), but it is compared against `receiving_vector.size()` which returns `size_t` (unsigned).

If `ePlayer` is negative (e.g. `NO_PLAYER = -1`), C++ implicit conversion rules convert it to unsigned before comparison: `(size_t)-1` becomes `0xFFFFFFFF` (4,294,967,295 on 32-bit).

This causes two failures:

1. **Line 58:** `ePlayer < receiving_vector.size()` → `4294967295 < size` → always **false**, so the code falls into the `else` branch.
2. **Line 66:** `while (ePlayer >= receiving_vector.size())` → `4294967295 >= size` → always **true**, so the loop attempts to grow the vector to ~4 billion entries (`push_back(NULL)` in an infinite-for-practical-purposes loop).

On a 32-bit process limited to ~3.5 GB address space, this immediately exhausts memory and crashes the game.

## Security Impact

**Severity: Critical — Remote Denial of Service in Multiplayer**

The `CvNetwork::receive()` function is called by the closed-source exe's network layer when a packet arrives from another player. If the exe passes the player ID from the packet without validation, a malicious player can send a single crafted network packet with an invalid player ID to crash the game for **all players** in the multiplayer session.

Even if the exe does validate, defensive coding requires that the DLL not have a crash-on-invalid-input path in its network handler.

## Fix

Add a bounds check at the top of `CvNetwork::receive()`:

```cpp
if (ePlayer < 0 || ePlayer >= MAX_PLAYERS) return;
```

This is a one-line fix with no gameplay impact — it simply rejects invalid player IDs before they can trigger the unsigned conversion bug.

acato:
## Summary
- Adds a bounds check at the top of `CvNetwork::receive()` to reject invalid player IDs (`< 0` or `>= MAX_PLAYERS`) before they reach the signed-to-unsigned comparison with `vector::size()`
- Without this check, a negative `ePlayer` (e.g. `NO_PLAYER = -1`) is implicitly cast to `~4 billion` when compared to the unsigned `size_t`, causing an infinite `push_back` loop that exhausts the 32-bit address space and crashes the game

Fixes #1225

## Test plan
- [ ] Verify multiplayer games still function normally (valid player IDs unaffected by the early return)

### DIFF
--- a/Project Files/DLLSources/Network/CvNetwork.cpp
+++ b/Project Files/DLLSources/Network/CvNetwork.cpp
@@ -55,6 +55,11 @@ void CvNetwork::addToVector(const int (&data)[CvNetwork::iPacketSize])
 // upon receipt of a packet by the exe.
 void CvNetwork::receive(PlayerTypes ePlayer, const int (&data)[CvNetwork::iPacketSize])
 {
+	if (ePlayer < 0 || ePlayer >= MAX_PLAYERS)
+	{
+		return;
+	}
+
 	if (ePlayer < receiving_vector.size() && receiving_vector[ePlayer] != NULL)
 	{
 		// already receiving a multi part message


Schmiddie1979:
I reviewed this against the current implementation of `CvNetwork::receive()`, and the issue described in this PR is valid.

`ePlayer` is used both in comparisons against `receiving_vector.size()` and subsequently as an index into `receiving_vector`, without validating the player ID first. Since `PlayerTypes` can contain negative values such as `NO_PLAYER`, this can lead to invalid behaviour due to the signed/unsigned comparison and ultimately to an invalid vector access or uncontrolled growth of the vector.

Adding the proposed guard at the beginning of `CvNetwork::receive()`:

```cpp
if (ePlayer < 0 || ePlayer >= MAX_PLAYERS)
{
    return;
}
```

is therefore the correct place to handle this. It ensures that the remainder of the function can safely assume a valid player ID and does not change the behaviour for any valid player.

I also considered whether an additional assertion would be useful here, but I don't think it is necessary. If `NO_PLAYER` can legitimately reach this interface under certain circumstances, silently rejecting the invalid packet is preferable to generating an assertion in debug builds.

The change is also very limited in scope and isolated to this part of the DLL. I have therefore accepted the PR. If any unexpected issues should nevertheless appear later, the affected code path should be relatively easy to identify and the change can be revisited without much difficulty.

Overall, I consider the fix correct, useful and appropriately scoped.

PR #1232, opened 2026-04-03
Clamp enumByteSize read count to prevent global buffer overflow

### LINKED ISSUE #1229: Security: enumByteSize global buffer overflow via crafted savegame
## Bug Description

In `CvSavegame.cpp:162-168`, the savegame reader loads the `enumByteSize` table:

```cpp
byte iSize;
Read(iSize);
for (int i = 0; i < iSize; ++i)
{
    Read(enumByteSize[i]);
}
```

`enumByteSize` is a global `char[NUM_SAVEGAME_CLASS_TYPES]` array where `NUM_SAVEGAME_CLASS_TYPES = 23` (`CvSavegame.cpp:46`). But `iSize` is a `byte` (0-255). A crafted savegame can set `iSize` to any value up to 255, writing **232 bytes of attacker-controlled data** past the end of the array.

## How it can be exploited

`enumByteSize` controls how many bytes are read for *every subsequent tag* in the savegame (`CvSavegame.cpp:304`):

```cpp
iSize = enumByteSize[m_eClassType];
variable = (SavegameVariableTypes)ReadUnsignedBytes(iSize);
```

The adjacent globals in memory include `xmlEnumByteSize` and `conversion_table[NUM_JITarrayTypes]` (a `std::vector<short>` array).

**Attack scenario (multiplayer):**

1. Attacker crafts a savegame where the `enumByteSize` marker is followed by `iSize = 0xFF` (255).
2. The next 255 bytes overwrite `enumByteSize[0..22]` (the legitimate table) **plus 232 bytes of adjacent globals**.
3. By overwriting `enumByteSize` entries with chosen values, the attacker controls how many bytes are consumed for each savegame tag read. Setting a value to 4 instead of 2 (or vice versa) causes every subsequent tag to be misread — the reader becomes permanently misaligned.
4. This misalignment means attacker-controlled data bytes are now interpreted as vector lengths, allocation sizes, array indices, and enum values throughout the rest of deserialization.
5. The `conversion_table` vectors (adjacent in memory) have their internal `std::vector` bookkeeping (pointer, size, capacity) overwritten. When the reader later indexes into `conversion_table[eType]`, it dereferences an attacker-controlled pointer — a direct path to **arbitrary read/write** and potentially **arbitrary code execution**.

This affects every player who loads the crafted savegame in a multiplayer session.

## Fix

Clamp `iSize` to `NUM_SAVEGAME_CLASS_TYPES`:

```cpp
byte iSize;
Read(iSize);
if (iSize > NUM_SAVEGAME_CLASS_TYPES)
{
    iSize = NUM_SAVEGAME_CLASS_TYPES;
}
for (int i = 0; i < iSize; ++i)
{
    Read(enumByteSize[i]);
}
```

acato:
## Summary
- Clamps the `iSize` byte read from the savegame to `NUM_SAVEGAME_CLASS_TYPES` (23) before using it as a loop bound for writing into the `enumByteSize` global array
- Without this check, a crafted savegame sets `iSize` up to 255, overwriting 232 bytes of adjacent globals — including the conversion table that controls all subsequent deserialization

Fixes #1229

## Test plan
- [ ] Load an existing savegame — verify it loads identically to before
- [ ] Save and reload mid-game — verify no data loss

### DIFF
--- a/Project Files/DLLSources/CvSavegame.cpp
+++ b/Project Files/DLLSources/CvSavegame.cpp
@@ -162,6 +162,10 @@ CvSavegameReader::CvSavegameReader(CvSavegameReaderBase& readerBase)
 			// read the class enum sizes
 			byte iSize;
 			Read(iSize);
+			if (iSize > NUM_SAVEGAME_CLASS_TYPES)
+			{
+				iSize = NUM_SAVEGAME_CLASS_TYPES;
+			}
 			for (int i = 0; i < iSize; ++i)
 			{
 				Read(enumByteSize[i]);


Schmiddie1979:
I think there may be a small issue with clamping iSize directly to NUM_SAVEGAME_CLASS_TYPES.

iSize does not only control how many entries are written to enumByteSize; it also represents how many bytes are present in the savegame for this array. If a savegame contains an iSize larger than NUM_SAVEGAME_CLASS_TYPES, clamping it before the loop prevents the out-of-bounds write, but it also means that the remaining bytes are never consumed from the savegame stream. This could leave the reader at the wrong position and cause subsequent data to be interpreted incorrectly.

I think it would be safer to read all iSize entries, but only store those which fit into the current enumByteSize array:

```cpp
byte iSize;
Read(iSize);

for (int i = 0; i < iSize; ++i)
{
    char iByteSize;
    Read(iByteSize);

    if (i < NUM_SAVEGAME_CLASS_TYPES)
    {
        enumByteSize[i] = iByteSize;
    }
}
```

This still prevents writing past the end of enumByteSize, while ensuring that all bytes belonging to the saved array are consumed and the stream position remains correct. It should also behave better if a savegame from a newer version legitimately contains more savegame classes than the current DLL knows about.

PR #1234, opened 2026-04-03
Return -1 instead of raw index when ConvertIndex lookup fails

### LINKED ISSUE #1231: Security: ConvertIndex passthrough enables out-of-bounds EnumMap writes via crafted savegame
## Bug Description

In `CvSavegame.cpp:530-543`, when converting enum indices from a savegame:

```cpp
int CvSavegameReader::ConvertIndex(JITarrayTypes eType, int iIndex) const
{
    JITarrayTypes eBaseType = GetBaseType(eType);
    if (iIndex >= 0 && eBaseType >= 0 && eBaseType < NUM_JITarrayTypes)
    {
        if ((int)conversion_table[eBaseType].size() > iIndex)
        {
            return conversion_table[eBaseType][iIndex];
        }
    }
    return iIndex;  // <-- raw savegame index passed through unchanged
}
```

If the savegame contains an enum index that exceeds the conversion table's size, the raw index from the savegame is returned **unchanged**. This value is then used to index into EnumMap arrays, where the bounds check is `FAssert` — **a no-op in Release builds**.

## How it can be exploited

The conversion table maps old enum indices to current ones. Its size equals the number of entries that existed when the savegame was created. The attacker controls both the conversion table contents (written in the savegame header) and the enum indices used throughout the save.

**Attack scenario (multiplayer):**

1. Attacker crafts a savegame with a deliberately short conversion table for a given enum type (e.g., `YieldTypes` with only 5 entries instead of the expected 40+).
2. In the body of the save, the attacker uses an enum index of, say, 200 for that type.
3. `ConvertIndex` checks: `conversion_table[eBaseType].size() > 200` → false (table only has 5 entries).
4. The raw index 200 is returned unchanged.
5. This index reaches `EnumMap::set(200, value)` in `CvSavegameEnumMap.h`. The EnumMap may only have storage for 40 elements. Since `FAssertBounds` is a no-op in Release, the write proceeds out-of-bounds.
6. EnumMaps are embedded inline in game objects (`CvCity`, `CvUnit`, `CvPlayer`). An out-of-bounds write to an EnumMap **corrupts adjacent member variables** in the same object.
7. The attacker controls both the index (how far past the array to write) and the value written, enabling targeted corruption of specific fields: owner IDs, coordinates, hit points, gold — any field adjacent to an EnumMap in the object layout.

This is especially dangerous because EnumMaps are **pervasive** — they appear dozens of times in every major game object, each one a potential write primitive.

## Fix

When the conversion table lookup fails (index out of range), return -1 instead of passing through the raw index. The value -1 already has established semantics in the codebase as "no valid entry" and is handled safely by callers:

```cpp
// Index exceeds conversion table — treat as unknown/removed entry
return -1;
```

acato:
## Summary
- Changes `ConvertIndex()` to return -1 (the established "removed entry" sentinel) when a savegame enum index exceeds the conversion table size, instead of passing through the raw attacker-controlled index
- Without this, an out-of-range index passes through to `EnumMap::set()` where the bounds check (`FAssert`) is stripped in Release builds, enabling out-of-bounds writes into game object memory

Fixes #1231

## Test plan
- [ ] Load an existing savegame — verify it loads identically to before
- [ ] Load a savegame from an older version with different XML content — verify conversion still works correctly for valid indices
- [ ] Save and reload mid-game — verify no data loss

### DIFF
--- a/Project Files/DLLSources/CvSavegame.cpp
+++ b/Project Files/DLLSources/CvSavegame.cpp
@@ -539,7 +539,7 @@ int CvSavegameReader::ConvertIndex(JITarrayTypes eType, int iIndex) const
 		}
 	}
 
-	return iIndex;
+	return -1;
 }
 
 int CvSavegameReader::GetXmlSize(JITarrayTypes eType) const


Schmiddie1979:
Thanks for the fix. I took a closer look at `ConvertIndex()` and the surrounding savegame conversion logic.

The change makes sense to me.

The conversion table already explicitly uses `-1` when an XML entry stored in an older savegame no longer exists in the current XML:

```cpp
// Do note that -1 will be used if something is removed from xml
```

Therefore, if a saved index cannot be resolved because it is outside the conversion table, returning the raw index is potentially dangerous. At that point there is no reason to assume that the old numeric index is still a valid index in the current XML data.

Returning `-1` instead is consistent with the existing handling of removed XML entries and ensures that an invalid old index is treated as an unresolved entry rather than being passed on as if it were valid.

It also avoids the more serious issue described in the PR, where such an index could eventually reach `EnumMap::set()` and potentially result in an out-of-bounds write in builds where the `FAssert` does not protect against it.

The normal conversion path is unaffected:

```cpp
if ((int)conversion_table[eBaseType].size() > iIndex)
{
	return conversion_table[eBaseType][iIndex];
}
```

Only the failed lookup now returns the established invalid value instead of the unverified raw index.

Small, targeted and consistent with the existing savegame conversion design.

Thanks for the fix. Merging this.

Schmiddie1979:
Unfortunately I had to revert this change after testing it with existing savegames.

Returning `-1` in this fallback path causes widespread savegame loading issues and crashes. Apparently this path also covers cases where preserving the raw index is required, so it cannot be treated in the same way as an XML entry which has actually been removed.

The underlying concern may still be valid, but we will need a more targeted solution which distinguishes between these cases.

PR #1257, opened 2026-08-15
Fix city title-bar hover and expand yield/health tooltips

raphaelmonticello:
## Summary
City-screen title-bar stats that sit over the 3D city view (health, happiness, harbour, barracks, defense) were losing hover to terrain plot help. This PR moves those hoverable widgets off the map frame, keeps the remaining title-bar stats readable, and fills gaps in the yield/health tooltip breakdowns.

## City screen UI (`CvMainInterface.py`)
- The 3D city viewport steals mouse hover from `setLabel` widgets drawn over it. Hoverable title-bar stats now use `setText` + `Button_Stone_Style` (same pattern as Growing / population) and are placed left of the map or right of `iMapRight`.
- City scroll arrows sit next to the city name so they no longer collide with the left stats.
- Happiness stays after law/crime on the left; health, harbour, barracks, defense, and storage sit on the right.
- Culture header now shows surplus + stored/threshold + turns, e.g. `4♣ 58/200 (36 turns)`.
- `CityTopBackground` is hit-test solid so transparent title-bar art does not pass clicks through to the map.

## Yield / health tooltips (`CvGameTextMgr.cpp` + text XML)
- `setCityYieldModifierString` was only called when leftover plot `iBaseProduction != 0`. Indoor/building yields (hammers, bells, crosses, etc.) therefore skipped the percent breakdown, so Final Production could show 9 while the tooltip listed 7 produced. The modifier lines now appear whenever modified production is non-zero.
- Civilization `getYieldRateModifier` and tax `getTaxYieldRateModifier` are now listed in that breakdown.
- Health-by-location is no longer a single combined line. The tooltip lists each applied city-center bonus:
  - fresh water
  - coast
  - hills
  - bad terrain (with the actual terrain name)
- If none of those apply, the old `Health by city location = 0` line is kept.

## Test plan
- [x] Compiled FinalRelease DLL locally and verified the city screen in-game.
- [ ] Full-quit Colonization and reload so both Python and the DLL are picked up.
- [ ] Hover health / happiness / harbour / barracks / defense on the city title bar: breakdowns, not terrain help.
- [ ] Confirm city name, scroll arrows, Growing, left yields, and right storage/EXIT do not overlap at 1280 and 1920 widths.
- [ ] Hover hammers / bells / crosses on a city whose plot does not produce that yield: percent modifiers still appear and match Final Production.
- [ ] Hover health on cities with fresh water, coast, hills, and/or bad terrain: each applied bonus is listed separately.

## Notes
- DLL change: players need a rebuilt `CvGameCoreDLL.dll`. The compiled binary is gitignored and is not in this PR.
- Local compile helper `compile_finalrelease.bat` is not included.
- Happy to split the Python layout from the DLL tooltip work if you would rather review them separately.

### DIFF
--- a/Assets/Python/Screens/CvMainInterface.py
+++ b/Assets/Python/Screens/CvMainInterface.py
@@ -545,6 +545,12 @@ class CvMainInterface:
 			size = 4
 		return "<font=" + str(size) + ">" + Text + "</font>"
 
+	def setTitleBarHelpText( self, screen, name, szBuffer, x, widgetType, data1, data2 = -1 ):
+		# setText + stone style matches PopulationText. Use only at X positions outside the 3D city view.
+		screen.setText(name, "Background", szBuffer, CvUtil.FONT_CENTER_JUSTIFY, x, CITY_TITLE_BAR_HEIGHT / 12, -0.3, FontTypes.SMALL_FONT, widgetType, data1, data2)
+		screen.setStyle(name, "Button_Stone_Style")
+		screen.show(name)
+
 	# Will Initialize the majority of Background panels and Widgets
 	def interfaceScreen ( self ):
 		if (CyGame().isPitbossHost()):
@@ -572,6 +578,8 @@ class CvMainInterface:
 	# CITY BURLAP BACKGROUND PANELS
 		screen.addPanel("CityTopBackground", u"", u"", True, False, 0, 0, xResolution, CITY_TITLE_BAR_HEIGHT, PanelStyles.PANEL_STYLE_STANDARD, WidgetTypes.WIDGET_GENERAL, -1, -1 )
 		screen.addDrawControl("CityTopBackground", ArtFileMgr.getInterfaceArtInfo("INTERFACE_CITY_BG_TOP").getPath(), 0, 0, xResolution, CITY_TITLE_BAR_HEIGHT, WidgetTypes.WIDGET_GENERAL, -1, -1 )
+		# Title-bar art is transparent over the 3D city view, so default hit-test lets terrain steal hover.
+		screen.setHitTest("CityTopBackground", HitTestTypes.HITTEST_SOLID)
 		self.appendtoHideState(screen, "CityTopBackground", HIDE_TYPE_CITY, HIDE_LEVEL_ALL)
 
 		screen.addPanel("CityLeftBackground", u"", u"", True, False, 0, CITY_TITLE_BAR_HEIGHT, CITIZEN_BAR_WIDTH, yResolution, PanelStyles.PANEL_STYLE_STANDARD, WidgetTypes.WIDGET_GENERAL, -1, -1 )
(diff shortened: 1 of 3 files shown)


Schmiddie1979:
Regarding the **city-screen UI changes,** I am currently unable to reproduce the issue described in the PR.

I tested the current develop branch at several different resolutions, specifically checking the tooltips in the top bar (Health, Happiness, Harbour, Barracks, Defense, etc.). In all of my tests, the corresponding tooltips are displayed correctly. I do not get terrain tooltips instead, and changing the resolution does not appear to cause any problems with them.

Perhaps I am misunderstanding the exact issue you are trying to fix. Could you please provide one or two screenshots showing the problem before the changes from this PR are applied, and ideally another screenshot showing the result afterwards?

It would also be helpful if you could provide the exact resolution and display configuration where you can reproduce it.

As I understand the proposed fix, the issue is that the 3D city viewport intercepts the mouse hover for the existing labels, and the solution is therefore to change/reposition those UI elements. However, since I cannot reproduce this behaviour on the current develop branch at any of the resolutions I tested, I would like to understand the exact circumstances under which this happens before we change the existing city-screen layout.

If this turns out to be specific to a certain resolution, display configuration, or another particular condition, we should probably identify that first. Otherwise, we may be changing UI code which currently works correctly for most configurations in order to solve a problem whose actual cause lies somewhere else.

So for now, I would prefer to see a reproducible example and screenshots of the issue before merging this part of the PR.

I would actually appreciate screenshots for the other UI and tooltip changes in this PR as well.

Since most of these changes affect how information is presented to the player, it is quite difficult to evaluate their usefulness from the code diff alone. For each of these changes, it would be very helpful to see:

- the current behaviour on the develop branch,
- the behaviour with this PR applied,
- and a short explanation of the specific problem or missing information the change is intended to solve.

This would be particularly useful for the yield/production tooltip changes, the health breakdown, the culture header, and the changes to the city-screen top bar.

Some of these may well be useful improvements, but before changing an established UI it should be clear what the player-facing problem currently is and how the proposed change improves it. A simple before/after screenshot for each affected area would make reviewing this PR considerably easier.

raphaelmonticello:
UI Before: I'm hovering over the health, but it's showing the tooltip of the terrain behind it instead
ui-before
UI After: as I couldn't find a solution for the terrain tooltips, I moved the icons from outside the top part of the terrain (and here you can see the health with extra info): 
ui-after

The resolution I'm experiencing it is 1920x1080

Schmiddie1979:
Hm, strange. I do not have these issues - with all possible resolutions it works perfect for me - tooltips are shown correctly.

Schmiddie1979:
Ok. issue identified. 1-Plot City radius causes the issue, 2-Plot does not. 

Hence, we need a totally different approach. I think I know why this issue exists. City Screen fits for 2-Plot. 1-Plot came back after a while, but the radius for the field plots should have to do something with this issue. Let me try to figure this out, maybe its an easy solution. I have an idea...

Schmiddie1979:
Ok, fix uploaded to the develop branch. Commit rejected.

PR #1258, opened 2026-08-15
Native village teaching from local yields

raphaelmonticello:
1. native villages were teaching specialties that did not exist nearby (cotton with no cotton, farmer on a water-locked island).

2. a village only teaches from what is actually in its 2-plot ring. named crop/animal experts need a visible bonus. name match is by tokens, not substrings (GRAPES does not unlock Rapeseed Planter). same tribe does not repeat a specialty already taught by another of its villages.

3. adjacent tiles score 3x a tile two steps away. farmer / scout / prospector / miner / seal hunter / pearls hunter are tied to the matching bonus. hills and peaks do not teach miner or prospector by themselves. hunter/trapper may still use terrain. coastal and great-river villages may teach fisherman without a fish bonus.

4. existing saves keep their already-assigned specialty. new game required.

5. game-speed / gift-attitude / refuse-to-talk clocks moved to #1262 so both PRs share one AI_effectiveTurns helper.

6. needs a new DLL. the compiled binary is not in this PR.


Schmiddie1979:
**Native village teaching**
I like the general idea behind the **native teaching changes.** It certainly makes sense that, for example, an Inuit settlement should not teach a cotton planter if cotton is completely foreign to that region.

However, I am a bit concerned about making the surrounding plots the primary or decisive factor for selecting the profession. Native teaching also has an important map-balancing function: ideally, the native settlements on a map should provide a reasonably broad selection of professions. With a purely local approach, certain professions might become very common while others could become extremely rare or even disappear entirely, depending on map generation.

I think a better solution might be to combine several factors rather than replacing the existing balancing/randomness with local yield suitability:

Regional plausibility: professions which are clearly inappropriate for the wider region or climate should be excluded or receive a strong penalty. This would prevent cases such as Inuit settlements teaching cotton, sugar or tobacco planters.
Local suitability: nearby terrain, features and especially bonus resources could provide a significant positive weight. For example, cotton near a settlement should make a cotton planter considerably more likely, but the absence of cotton on the immediately surrounding plots should not necessarily make that profession impossible.
Global distribution / rarity: professions which are not yet taught, or are only taught by very few native settlements on the map, could receive an additional weight. Conversely, professions which are already widely available could receive less priority.
Some randomness: retaining a random component would avoid making the result completely deterministic and would help keep different maps interesting.

In simplified terms, the selection could eventually be based on something like:

score = regional suitability
      + local terrain/resource suitability
      + rarity / map-balancing bonus
      + random component

Before deciding on such an alternative, however, I think we should first take a closer look at how teaching professions are currently distributed across the map. I am not entirely sure whether the existing bestTeachUnitClass() logic, or another part of the native settlement generation code, already contains mechanisms intended to balance the availability or distribution of professions. If such balancing already exists, we should try to preserve it and add regional/local plausibility on top of it rather than replacing it.

It would therefore be useful to understand first whether profession selection is purely local to each settlement, or whether the existing code already takes into account which professions are taught elsewhere, directly or indirectly through its scoring and randomization.

I would therefore suggest moving this PR back to draft for now and taking another look at both the existing teaching algorithm and the proposed local-yield logic. The underlying idea is good, but I think we should make sure that local plausibility and global profession availability are balanced together, rather than optimizing only for the immediate surroundings of each settlement.



**Gift attitude and readout** **/ Counter**
Great catch. The issue with AI_getHasMetCounter() is real, and I think the proposed game-speed adjustment for this particular case is fundamentally correct.

However, while looking into this, I noticed that HasMetCounter is only one of several counters which are incremented by a fixed +1 per turn in CvTeamAI::AI_doCounter(). Before implementing a solution specifically for HasMetCounter, I think we should take a closer look at the other counters as well and determine whether this is actually part of a more general game-speed scaling issue.

The counters we found are:

- WarPlanStateCounter
- AtWarCounter
- AtPeaceCounter
- HasMetCounter
- OpenBordersCounter
- DefensivePactCounter
- ShareWarCounter

All of these essentially count raw game turns. This is not necessarily wrong by itself. The important question is how each counter is used afterwards. Some mechanics may intentionally be turn-based, while others represent a duration which should probably remain comparable across different game speeds.

There are also indications that this has already been taken into account in some places. For example, some uses of WarPlanStateCounter in AI_doWar() already apply game-speed-dependent modifiers. Therefore, changing the counters themselves globally would probably not be a good idea, as this could result in double scaling for mechanics which already compensate for game speed.

On the other hand, some other uses look worth investigating. In particular, DefensivePactCounter and ShareWarCounter are used with a fixed threshold of 40 turns for Permanent Alliance considerations, and there are also fixed thresholds involving AtWarCounter. These may potentially have the same underlying problem as the HasMetCounter calculation.

One possible approach would therefore be to keep all counters as raw turn counters and provide a common helper for normalizing a counter when it is used by a mechanic which should scale with game speed. The actual scaling factor could still depend on the mechanic. For example, GrowthPercent seems reasonable for long-term diplomatic relations such as HasMetCounter, DefensivePactCounter and ShareWarCounter, while military planning may be better related to TrainPercent — which is already done in some places.

So I think the proposed change for the trade/gift attitude calculation is going in the right direction, but we may have discovered a broader issue here. It would be useful to audit the uses of these counters first and classify them into:

already game-speed adjusted,
should probably be game-speed adjusted,
intentionally based on raw turns.

I would therefore suggest moving the PR back to draft for now and investigating this a little further before implementing a singular solution for HasMetCounter. We may either find that the other counters are already handled correctly, or we may be able to develop a cleaner common solution for all affected cases.

raphaelmonticello:
about your comment on "Native village teaching":
1- the solution already resolves the situation of regional plausability. The original method is actually quite bad, being common for tribes to teach something that makes no sense.
2- the factor of randomization and balancing still persists from original to some degree. We try to not repeat the same teaching in the same tribe.
3-The method proposed is after multiple trials I did locally, the current method is creating much more interesting maps that it was before 

about the "Gift attitude and readout / Counter": great comments, I'll review it all and study changes and do some tests before I comment more about it.

raphaelmonticello:
about your comments — I reviewed the counters and did more local tests. I just pushed an update.

**Native village teaching**
I did not add the global rarity/random layer yet. After more playtests I still prefer the local method for now, it is creating more interesting maps. Same-tribe no-repeat is still only a tie-breaker. If after more maps we still see too many of one profession we can add your rarity term then.

**Gift attitude / Counter**
Your approach is the right one. I audited the counters and left them as raw +1 in AI_doCounter.

New helper:
- `AI_effectiveTurns(raw)` uses GrowthPercent
- `AI_effectiveTurns(raw, percent)` when the mechanic should use TrainPercent

Applied:
1- gift/trade attitude: Growth helper
2- Permanent Alliance 40-turn check: Growth
3- peace-deal war length and the 10/20/30 peace thresholds: Train
4- refuse-to-talk: Train
5- the short tactical AtWar < 10 check I left raw

Also I added remaining turns until peace talks in the refuse-to-talk hover/dialog, because that clock is now speed-scaled and otherwise invisible. Gift bonus stays blue, decay and peace-talk turns yellow. The gift readout now shows for every leader, not only natives.

Still draft. I'll keep testing.

Schmiddie1979:
Ok, thanks. Again: Before making any commits to the develop branch it would be great to read your ideas on DC - especially if the changes provide deeper changes regarding core features of the mod!

raphaelmonticello:
about the latest teaching tests:

1- I kept iterating on the local-yield method after the first maps. The current version is what I have been playtesting.

2- named experts now need the bonus icon you can actually see. grapes no longer turn a village into rapeseed planter (that was a substring match). same for seals vs capybara and pearls vs gems.

3- same tribe does not repeat a skill when another local option exists.

4- adjacent tiles count more than 2 tiles away, so a mine across the water loses to something next to the village.

5- fur/hunter is scored lower, so rarer goods (pepper, horses -> scout, etc.) show up more. I was seeing too many hunters.

I am quite happy with the current version in the tests I am doing. maps look much more interesting than before. happy to tweak if something still looks off.

Schmiddie1979:
How does the gift attitude / counter part work? Have you tested this part as well? I currently wonder if we could separate the village teaching part from the speed-scaled thing?

RaphaelDeMonticello:
> How does the gift attitude / counter part work? Have you tested this part as well? I currently wonder if we could separate the village teaching part from the speed-scaled thing? 

Yes, the village teaching can be separated.

The the gift/counter part is as described in the PR. It's being useful for the player on the games I playtested

raphaelmonticello:
1. Schmiddie asked to split teaching from the general game-speed clocks.
2. this PR is now teaching only (CvCity::bestTeachUnitClass).
3. gift attitude, refuse-to-talk, peace clocks, and the shared AI_effectiveTurns helper moved to #1262.

Schmiddie1979:
I will work on this topic now...

Schmiddie1979:
Thanks for the work on this. I have now done a first round of testing with the current version of the PR.

From the technical side, the implementation looks well thought out. The new logic is deterministic, the local plots are evaluated in a structured way, and I especially like the idea of taking the surroundings of a native village into account instead of relying mainly on the previous random selection. The attempt to avoid duplicate teaching specialities within the same tribe is also a good idea in principle.

Unfortunately, the first in-game results are not balanced yet. On larger test maps, the distribution of teaching specialities becomes very repetitive. Miners/Prospectors in particular appear extremely often, followed by a relatively small group of other common specialities such as fishermen and fruit gatherers. At the same time, several of the classic and more valuable/interesting specialities – for example tobacco planters, cotton planters and similar crop experts – were either extremely rare or did not appear at all in my tests.

The duplicate prevention within a tribe does help locally, but it does not really solve this problem on the map as a whole. A tribe may avoid teaching the same profession twice, but the next tribe then tends to generate a very similar set again. The practical result can therefore be something like Miner/Prospector + Fisherman + Fruit Gatherer for several different tribes, while some of the traditional crop specialists are completely missing from the map.

Looking at the code after seeing these results, I think there are probably two effects involved. First, some special cases receive very strong scores – especially the precious-metal/Prospector logic – which can make them dominate the selection. Second, the matching of named crop/animal specialists to local bonuses relies quite heavily on the XML type names and string matching. This is clever and keeps the implementation local, but we should check whether all existing WTP resources and specialists are actually matched reliably. The absence of specialists such as Tobacco Planters or Cotton Planters in the tests makes me suspicious that some of these candidates may not receive the intended local bonus at all.

So my first impression is that the technical approach is promising, but the resulting distribution still needs substantial balancing before I would merge it. I would not simply reduce the prospector score, because that may only hide one symptom. I think we should first verify that all important resource → specialist combinations are detected correctly, and then reconsider the relative weights.

Ideally, local geography should strongly influence what a village teaches without making the overall distribution too predictable. A village near tobacco should have a very good reason to teach a Tobacco Planter, gold/silver should make a Prospector more likely, an arctic coastal village should naturally favour appropriate professions, etc. But across a complete map there should still be enough variety that rare and characteristic specialists do not disappear while every tribe ends up offering essentially the same small group of professions.

In short: the code itself looks clean and the concept is good, but the first gameplay tests show that the selection is currently much too concentrated on a few professions. I think this is absolutely worth continuing to work on, because getting the balance right could make native village specialities much more interesting and geographically meaningful than the old system.

From a gameplay perspective, I would also put more emphasis on professions that feel natural for native villages: hunters, gatherers, farmers and planters should make up the majority of the teaching specialities. Miners and Prospectors can certainly appear, and they are very attractive for the player, but they should probably remain comparatively rare and tied to genuinely exceptional local conditions.

At the moment, the balance seems almost reversed in some test cases: mining-related specialists appear very frequently, while several of the classic agricultural and gathering professions are missing or strongly underrepresented. That makes the system feel less geographically varied than intended, even though the underlying logic is trying to achieve exactly that.

Ideally, I would expect the common baseline to come from hunting, gathering and agriculture, with crop-specific planters appearing where the local resources justify them. Mining professions should then be the special case rather than one of the dominant outcomes. That would probably also make the rare mining villages feel much more distinctive when the player actually finds one.



Another issue I noticed in the first tests concerns specialities that are not directly tied to a specific resource. Scouts are a good example. I did not encounter Scouts at all in my initial test maps.

I think we need to be careful that the stronger focus on local yields and resources does not unintentionally eliminate professions whose role is not based on a particular map bonus. These professions still need a reasonable way to enter the selection. Otherwise the new system may improve the geographical logic for resource-based professions while at the same time removing part of the variety that existed before.

So I would probably distinguish between two groups: professions that should be strongly influenced by local resources and terrain, and professions such as Scouts that need some other weighting or fallback mechanism because there is no corresponding resource that would naturally trigger them.

raphaelmonticello:
> At the same time, several of the classic and more valuable/interesting specialities – for example tobacco planters, cotton planters and similar crop experts – were either extremely rare or did not appear at all in my tests.

If this happened, it means no bonus of tobacco/cotton were near the village. so, on old logic, it would appear on a random place far away that made no sense (as such as the case)
Tobacco/Cotton (and other)  planters appears normally on my tests when there are these bonus resources around it

> Looking at the code after seeing these results, I think there are probably two effects involved. First, some special cases receive very strong scores – especially the precious-metal/Prospector logic – which can make them dominate the selection

The reason to prioritize precious-metal/prospectos when it's nearby is because those are rare, so to avoid them becoming miners instead.

> From a gameplay perspective, I would also put more emphasis on professions that feel natural for native villages: hunters, gatherers, farmers and planters should make up the majority of the teaching specialities. 

Hunters and there other professions also appear a lot, again, it depends if there are these bonus resources around the village (2 tiles)

---
One suggestion: this logic can also be a (selection button) on the Custom Game/Custom Scenario, so the player can chose between this logic or the old random one

Schmiddie1979:
Understood, but maps generate a lot of hills and peaks - therefore in my tests every 3rd village has miner and prospector - thats a real issue, because on a huge map player finds 50x miner, but no scout, no fisherman, no cotton planter etc. Maybe we should not combine it with bonus ressources, but only suitable plots?

Schmiddie1979:
Don't get me wrong, I like the feature, but currently it simply seems to produce weird results that will make a player mad

raphaelmonticello:
> t maps generate a lot of hills and peaks - therefore in my tests every

it's not about hills or peaks, it's about the bonus resources there.

In case it's not correct yet, we should fix it

raphaelmonticello:
> Don't get me wrong, I like the feature, but currently it simply seems to produce weird results that will make a player mad

I understand.
I believe if you are finding some unexpected results, feel free to tweak the logic.

Schmiddie1979:
It is, because also if there is no bonus ressource, village has miner or prospectors a lot...

raphaelmonticello:
you were right, it was the hills.

empty hills always give ore, so miner was winning with no bonus on the tile. now miner needs iron / bog ore / minerals / salt nearby. prospector still needs gold / silver / gems.

coastal and great river villages can teach fisherman without a fish bonus, so they don't all become hunter. cotton and scout still need the matching resource in the 2-plot ring.

pushed new commit.  needs a new game and a new DLL

Schmiddie1979:
I still like the local-environment approach very much, but after testing I think bonus resources currently have too much influence. Rare professions can disappear completely if their bonus isn't near a village, while Prospectors are still much too common.

My suggestion would be a weighted profession pool:

surrounding land/yield potential should be the main factor;
matching bonus resources increase the weight, but don't determine eligibility;
each profession gets a base weight controlling its general rarity;
on Huge+ maps, every regular teaching profession should ideally appear at least once, provided there is a reasonably suitable village.

As a first balancing proposal:

100–80: Farmer 100, Fisherman 90, Hunter 90, Fruits Picker 80
65–50: Trapper 65, Rice Farmer 60, Cassava Planter 60, Tobacco 55, Cotton 55, Sugar 55, Coffee 50, Cocoa 50, Hemp 50, Linen/Flax 50
45–35: Indigo 45, Peanut 45, Yerba 40, Grapes 40, Olives 40, Rape 35, Coca 35
30–20: Logwood 30, Cochineal 30, Red Pepper 30, Vanilla 30, Kautschuk 30, Maple Syrup 25, Scout 25, Seal Hunter 20, Pearl Hunter 20
10: Prospector

Local suitability could then modify these weights, e.g. unsuitable ×0, weak ×0.5, suitable ×1, good ×1.5, very good ×2. A matching bonus could add roughly +25–50% instead of dominating the score.

This should keep the main benefit of the PR — villages teaching professions fitting their environment — while preserving profession diversity and making rarity controllable independently from random bonus placement.

Happy to discuss.

RaphaelDeMonticello:
Implemented the weighted pool from your last comment.

- Surrounding land/yield is the main score (unsuitable x0, weak x0.5, suitable x1, good x1.5, very good x2).
- A matching bonus no longer decides eligibility. It only adds +25% at range 2 or +50% if adjacent.
- Base rarity weights match your table (Farmer 100 ... Prospector 10). Miner is 20 so hills can teach it from ore yield without flooding the map.
- Villages pick from that pool by weight. Same tribe still avoids repeating a specialty when another local option exists.
- On Huge / Massive / Gigantic, a still-missing regular profession is assigned first if the village is at least suitable for it.

Needs a new game and a new DLL (binary not in the PR).

Schmiddie1979:
great, will have a look tomorrow. thanks

raphaelmonticello:
I just tested it and I'm loving it.

Schmiddie1979:
Ok, great. seems to work nicely. 

Small note: UNITCLASS_PEAT_CUTTER might have been overlooked in nativeTeachBaseWeight(). It is available through the Native TeachUnitClasses and is used as the expert unit for both the Clay Miner and Peat Cutter professions, but currently falls back to the default weight of 40. Maybe it should get an explicit base weight like the other teachable specialists.

RaphaelDeMonticello:
Good catch. UNITCLASS_PEAT_CUTTER now has an explicit base weight of 30 (same band as the other collectors). It is the expert for both Peat Cutter and Clay Miner, so it was previously falling back to 40.

Needs a new game and a new DLL.

Schmiddie1979:
> I just tested it and I'm loving it.

Ok, seems to work fine now. Sometimes I had the impression that some professions might be not on the map, but that is only because I searched for it. Considering the former system - this is nothing new and part of the feature.

Schmiddie1979:
Thanks again for this great idea. I have taken over the idea and implemented it into develop with some additional improvements and balancing.

For development reasons I did not merge this PR directly, but the feature itself has been incorporated into develop.

PR #1260, opened 2026-08-17
Keep World Builder bonus placements when playing a scenario

raphaelmonticello:
## Summary

World Builder maps already store each plot bonus (`BonusType=BONUS_HEMP` etc.) and write `Randomize Resources=False` so Play Scenario is supposed to keep those placements.

`applyMap()` compared the flag only to lowercase `false`. The saved value is `False`, so the check treated every WB map as randomize, wiped all bonuses, and ran `addBonuses()` again. Features had the same case bug.

## Change

- Treat `Randomize Resources`, `Randomize Features`, and `Randomize Goodies` case-insensitively (`true` / `1` / `yes` randomize; anything else keeps the WB placements).
- Write the three flags as lowercase `false` on save so they match the stored default.
- Also write `Randomize Goodies=false` (it was read but never saved).

Python only. No DLL rebuild. Existing WB maps with `Randomize Resources=False` keep their bonuses after a game restart.

## Test plan

- Save a World Builder map with a few hand-placed bonuses.
- Play Scenario from that map: those bonuses stay on the same tiles.
- Optional: set `Randomize Resources=true` in the `BeginMap` block and confirm bonuses are regenerated.

## Notes

- No compiled `CvGameCoreDLL.dll`.
- `Randomize Resources=true` still forces a fresh scatter if a scenario author wants that.

### DIFF
--- a/Assets/Python/EntryPoints/CvWBInterface.py
+++ b/Assets/Python/EntryPoints/CvWBInterface.py
@@ -118,20 +118,31 @@ def getPlayerData():
 	"after reading a save file, return player data as a tuple, terminated by -1"
 	t=()
 	for i in range(gc.getMAX_CIV_PLAYERS()):
-		leaderType = CvUtil.findInfoTypeNum(WBDesc.playersDesc[i].leaderType)
-		civType = CvUtil.findInfoTypeNum(WBDesc.playersDesc[i].civType)
-		handicapType = CvUtil.findInfoTypeNum(WBDesc.playersDesc[i].handicap)
-		color = CvUtil.findInfoTypeNum(WBDesc.playersDesc[i].color)
-		artStyle = gc.getInfoTypeForString(WBDesc.playersDesc[i].artStyle)
+		pWBPlayer = WBDesc.playersDesc[i]
+		# Church / animals: the DLL creates these. Passing them in as scenario
+		# civs crashes initEuropeSettler because they have no cities.
+		if pWBPlayer.civType == "NONE" or CvWBDesc.isDllCreatedCiv(pWBPlayer.civType):
+			civType = -1
+			leaderType = -1
+			color = -1
+			artStyle = -1
+			isPlayableCiv = 0
+		else:
+			leaderType = CvUtil.findInfoTypeNum(pWBPlayer.leaderType)
+			civType = CvUtil.findInfoTypeNum(pWBPlayer.civType)
+			color = CvUtil.findInfoTypeNum(pWBPlayer.color)
+			artStyle = gc.getInfoTypeForString(pWBPlayer.artStyle)
+			isPlayableCiv = pWBPlayer.isPlayableCiv
+		handicapType = CvUtil.findInfoTypeNum(pWBPlayer.handicap)
 
 		t=t+(civType,)
-		t=t+(WBDesc.playersDesc[i].isPlayableCiv,)
+		t=t+(isPlayableCiv,)
 		t=t+(leaderType,)
 		t=t+(handicapType,)
-		t=t+(WBDesc.playersDesc[i].team,)
+		t=t+(pWBPlayer.team,)
 		t=t+(color,)
 		t=t+(artStyle,)
-		t=t+(WBDesc.playersDesc[i].isMinorNationCiv,)
+		t=t+(pWBPlayer.isMinorNationCiv,)
 
 	return t
 
--- a/Assets/Python/pyWB/CvWBDesc.py
+++ b/Assets/Python/pyWB/CvWBDesc.py
@@ -23,6 +23,12 @@ def getPlayer(idx):
 		return gc.getPlayer(idx)
 	return None
 
+def isWbFlagTrue(value):
+	# WB files write False/True; applyMap used to compare only to "false".
+	if value is None:
+		return False
+	return str(value).strip().lower() in ("true", "1", "yes")
+
 #############
 class CvWBParser:
 	"parser functions for WB desc"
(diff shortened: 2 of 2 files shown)


Schmiddie1979:
Sounds interesting. However, as a supporter of the mod - could you please share any ideas you like to change in DC before you commit them. Thanks.

raphaelmonticello:
oops, sorry, I thought that the PRs would count as the ideas implemented and tested already.
a few seconds ago I also pushed 2 other proposals PRs. 
Where do you want me to send the proposals first before implementing? I think it would slow things down, but I want to comply with the process you have stablished.

by the way, this one specifically seems like a bug and this PR fixes it

Schmiddie1979:
It will definitely slow things down, but this is not an open source project- ideas that change things of the mod need to be discussed or at least presented first in Discord so that we can have a look what shall be changed. Since the core team that knows the features and history of the mod is currently small uploads could unintendedly change balanced core features or at least features that shall not be changed.

Schmiddie1979:
https://discord.gg/UWSw2x5wqh

raphaelmonticello:
ok, just joined, I said a hello there.

Schmiddie1979:
This looks like a good and sensible fix in principle, and I would be happy to accept it.

However, there is one potentially important side effect that I think we should check before merging this.

Existing WorldBuilder maps may already contain:

Randomize Resources=False
Randomize Features=False

The current WorldBuilder code writes these values automatically. Due to the case-sensitivity bug, "False" is currently treated differently from "false", which means that resources/features may have been randomized when starting these maps despite the flags saying False.

After this fix, False would finally be interpreted correctly. This also means, however, that existing maps could suddenly start using the resource and feature placements stored in the WB file instead of generating them randomly as they did before.

In other words, some existing maps may have unintentionally relied on this bug for their current behaviour. The False value in an existing map does not necessarily prove that the map author deliberately wanted fixed resource placement, because the WorldBuilder itself writes this value when saving a map.

I therefore think we should test some of our existing maps (especially the scenario maps) before merging this, especially maps where resources are expected to be randomized when starting a new game - not sure if the mapper had this in mind. We should check whether they contain Randomize Resources=False / Randomize Features=False and whether their behaviour changes with this fix.

If some maps are supposed to keep randomizing resources, we may simply need to update those maps to explicitly use Randomize Resources=true. The fix itself still looks correct, but we should make sure that we do not unintentionally change the established behaviour of existing maps.

However, if this fix will have an impact on existing maps, we maxbe need to update all maps in order to get them working correctly resp. as intended.

raphaelmonticello:
1. I checked all 20 shipped WorldBuilder maps in PrivateMaps
2. they all already have Randomize Features=true, Randomize Resources=true and Randomize Goodies=true
3. so this fix does not change those maps. they already randomize today and they will keep randomizing
4. the only maps that change are player-made WB saves with False (the capital F from the old writer). those will finally keep the bonuses stored in the file, which is what that flag was meant to do

Schmiddie1979:
I am currently running into a separate WorldBuilder issue while trying to test this properly.

When I save a map in WorldBuilder and then load it as a scenario, I aleways get a traceback during `CvWBDesc.applyInitialTeams1()`, apparently while restoring team relations. This also happens on current `develop`, so it does not seem to be caused by this PR.

Since you mentioned that you regularly use the WorldBuilder: could you briefly describe your normal workflow for creating and reloading a WB map/scenario for testing?

In particular, do you:

- start a normal game and save it from WorldBuilder,
- start from an existing `.ColonizationWBSave`,
- or use some other workflow?

I want to make sure I am reproducing the resource issue under the same conditions you are using.

Schmiddie1979:
Tested this quite extensively now and I can confirm the fix.

The important part was comparing maps created on develop with maps created with this PR.

On current develop, a WorldBuilder map is saved with:

Randomize Resources=False

However, the loader checks against lowercase "false". As a result, a map created on develop does not preserve the manually placed resources when it is loaded as a scenario; the resources are randomized again.

With #1260 applied, a newly saved WorldBuilder map preserves the manually placed resources correctly.

I also did an additional cross-check which makes the result particularly clear: I created and saved a map with #1260, switched back to plain develop, and loaded that same map there. The resources were still preserved. This confirms that the relevant difference is already in the WorldBuilder file written by the PR (false instead of False), and that the existing develop loader then handles it correctly.

So the test results are:

map created on develop → manually placed resources are randomized when loaded;
map created with #1260 → manually placed resources are preserved;
map created with #1260 and subsequently loaded on plain develop → resources are also preserved.

I encountered a separate traceback during some of the scenario-loading tests, which initially made the results somewhat confusing. After additional testing I can reproduce that on plain develop as well, so it is unrelated to this PR and should be investigated separately.

#1260 works as intended. However, not merged so far. maybe these new insights help

raphaelmonticello:
1. thanks for the extra testing. the resource flag fix is unchanged
2. I looked at the separate traceback in applyInitialTeams1. WB files store ContactWithTeam/AtWar with the team itself, and a live-game save can also restore war vs a parent (that starts a revolution during load) or open borders/pacts when a team has no leader yet
3. scenario load now skips self-relations, king/colony declareWar, and open borders/pacts with no leader. it also stops walking off the end of the team list
4. python only, restart the game. my usual test path is: start a normal game, open WorldBuilder, save the map, then Play Scenario with that file

Schmiddie1979:
I retested the updated branch, but unfortunately the scenario still crashes when started normally.

Before the crash I now get these asserts:

CvGlobals.cpp, line 1324
CvGlobals::getCivilizationInfo
Expression: eCivilizationNum > -1

followed by:

CvPlayer.cpp, line 1594
CvPlayer::initEuropeSettler
Expression: is(CIV_CATEGORY_COLONIAL)

So the additional checks in applyInitialTeams1() do not seem to solve the underlying problem yet. It looks as if scenario initialization still reaches initEuropeSettler() for a player that is either not assigned a valid civilization or is not a colonial player.

The map still crashes when started as a normal scenario.

raphaelmonticello:
1. the applyInitialTeams1 guards only stopped the team-relations traceback. the crash you are seeing now is later, during scenario start

2. a live-game WB save writes Church and Wild Animals as players. official maps leave those slots as NONE and the DLL creates them in setInitialItems. those two civs are not playable, they have no cities, and they are not colonial, so Play Scenario later hits getCivilizationInfo(-1) and initEuropeSettler

3. load now treats Church / Wild Animals as empty slots (same as official maps), so the DLL can create them. write does the same for new WB saves. also RandomStartLocation=False is read case-insensitively, because that was the same True/False bug as the resource flag

4. python only, restart the game. you can retest with the WB file you already have, no need to save a new one. start it as a scenario as before

Schmiddie1979:
Retested the latest changes, including normal scenario loading. Everything is working correctly now and the previous crash is gone.

I also confirmed that resources can still be randomized explicitly when desired. Merged, thanks for the fix!

PR #1261, opened 2026-08-17
Show trade-quest standings and stolen crime gold

raphaelmonticello:
## Summary
Two small player-facing readouts that were missing numbers:

1. First-to-ship trade quests only showed the target amount, not who was ahead.
2. High crime removed gold but the city log only said gold was stolen.

## Trade-quest standings (`CvRandomEventInterface.py`)
- Europe, Africa, and Port Royal "first to ship / buy X" quest help now lists each living colonial civ.
- Exports use `getYieldSoldTotal`. Trade goods and luxury goods use `getYieldBoughtTotal`.
- The current leader is marked. Your own line is labeled You.

## Crime gold (`CvCity::doCityCrime`)
- The stolen amount (`iCityCrime`) is passed into `TXT_KEY_CITY_GOLD_STOLEN_BECAUSE_CRIME`.
- English/French/German/Russian/Hungarian texts show `%d2[ICON_GOLD]`.

## Test plan
- [x] Change is Python + one DLL text argument. Compiled FinalRelease locally with this plus other playtest work.
- [ ] Start a first-to-ship Europe/Africa/Port Royal quest: hover/help lists standings.
- [ ] Trade goods / luxury goods quests show bought totals, not sold.
- [ ] High-crime city: log shows how much gold was stolen.
- [ ] Existing saves: no save-format change.

## Notes
- Crime line needs a rebuilt `CvGameCoreDLL.dll`. The compiled binary is gitignored and is not in this PR.
- Trade standings apply after a game restart (Python).
- Local compile helper `compile_finalrelease.bat` is not included.

### DIFF
--- a/Assets/Python/CvEventManager.py
+++ b/Assets/Python/CvEventManager.py
@@ -426,6 +426,7 @@ class CvEventManager:
 		'Called at the beginning of a players turn'
 		iGameTurn, iPlayer = argsList
 		CvRandomEventInterface.checkEuropeanBorderDisputeKingWarDemand(iPlayer)
+		CvRandomEventInterface.checkTradeQuestLeadershipChanges(iPlayer)
 
 	def onEndPlayerTurn(self, argsList):
 		'Called at the end of a players turn'
--- a/Assets/Python/EntryPoints/CvRandomEventInterface.py
+++ b/Assets/Python/EntryPoints/CvRandomEventInterface.py
@@ -12397,17 +12397,15 @@ def CanDoEuropeTrade(argsList, iYieldID, iQuantity):
 		return False
 	return True
 
-def _appendTradeQuestStandings(szHelp, iYield, eLocation, eActivePlayer):
-	# Live race standings for first-to-ship-X quests. Sold totals for cargo exports;
-	# bought totals for trade/luxury goods imported from that market.
+def _getTradeQuestStandings(iYield, eLocation):
+	standings = []
 	if iYield < 0:
-		return szHelp
+		return standings
 
 	iTradeGoods = gc.getInfoTypeForString("YIELD_TRADE_GOODS")
 	iLuxuryGoods = gc.getInfoTypeForString("YIELD_LUXURY_GOODS")
 	bBuy = (iYield == iTradeGoods or iYield == iLuxuryGoods)
 
-	standings = []
 	for iPlayer in range(gc.getMAX_PLAYERS()):
 		loopPlayer = gc.getPlayer(iPlayer)
 		if loopPlayer is None or loopPlayer.isNone() or not loopPlayer.isAlive():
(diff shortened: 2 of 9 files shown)


Schmiddie1979:
The first change, showing the amount of gold stolen by city crime, looks very useful to me. It provides the player with relevant information about something that has already happened, without changing the underlying gameplay. I would be happy to accept this part.

I would prefer not to include the live standings for the trade quests, however. This is not a technical concern with the implementation, but a gameplay/design decision. Part of the appeal of these quests is that the player does not know exactly how much progress the other colonial powers have made. This uncertainty makes the race more interesting and creates some tension about whether another player might complete the objective first.

Showing the exact progress of all competitors would remove much of that uncertainty. Therefore, I would prefer to keep the current behaviour and not expose the other players' progress.

So from my side: the city crime gold display is welcome, but I would not include the trade quest standings.

raphaelmonticello:
@Schmiddie1979 for the ranking, what about just keep the amount the player already transported, so at least the player knows this personal info instead of having to keep track manually from multiple items. 
After some involvement in a match, the play may totally forger that it's missing just a few more to complete that quest, for example

Schmiddie1979:
Hm, I thought about that and changed my mind. It could be interesting to know who currently leads in order to take actions against this oponent....

raphaelmonticello:
So, do you want a list with I fly who leads (if not the player) and the amount shipped by the leader and the player?
Or only the name of who leads?(With  the amount shipped by the player)

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Schmiddie1979:
One question regarding the Africa trade quest: I noticed that its condition differs from the corresponding Europe and Port Royal conditions:

```cpp
elif iYield == gc.getInfoTypeForString("YIELD_TRADE_GOODS") or iYield == event.getGenericParameter(2) != gc.getInfoTypeForString("YIELD_LUXURY_GOODS"):
```

while Europe and Port Royal use:

```cpp
elif iYield == gc.getInfoTypeForString("YIELD_TRADE_GOODS") or iYield == gc.getInfoTypeForString("YIELD_LUXURY_GOODS"):
```

Is this difference intentional, or is the Africa condition an existing typo/bug?

I am asking because the new standings logic itself treats Trade Goods and Luxury Goods consistently for all three trade locations, so I want to make sure that the Africa quest text and the standings use the same logic.

raphaelmonticello:
1. that Africa line is an old typo, not something I added for the standings. I'll fix it
2. iYield is already getGenericParameter(2), so when the quest is luxury goods the condition becomes false and the buy-help text never shows
3. Europe and Port Royal already use TRADE_GOODS or LUXURY_GOODS. I’ll make Africa the same so the help text and the standings use the same logic

For the ranking, the current version lists every colonial civ, marks who leads, and shows your amount. If you want it smaller I can show only the leader (name + amount) plus you.

raphaelmonticello:
1. that Africa line is an old typo, not something I added for the standings
2. iYield is already getGenericParameter(2), so when the quest is luxury goods the buy-help text never shows
3. Europe and Port Royal already use TRADE_GOODS or LUXURY_GOODS. I made Africa the same so the help text and the standings use the same logic

raphaelmonticello:
sorry, I tested the other one abou the diplomacy. I'll test this one now

raphaelmonticello:
1. latest change: if you already have one of these first-to-ship quests, and the leader of that product changes, you get an event-log message
2. green if you take the lead, red if you lose it, yellow if it changes between other nations
3. it does not spam the turn you first get the quest. after that, the first real leader from everyone at 0 also counts
4. quest help still lists all nations, as we have now
5. python + xml only, just restart. texts EN/FR/DE/RU. no dll for this part

raphaelmonticello:
this is ready for review

Schmiddie1979:
Retested the latest changes. The trade quest display now updates automatically when the relevant goods change, so the issue I encountered before is fixed.

I wasn't able to specifically test the stolen crime gold change because I couldn't trigger a suitable crime event during testing, but the other changes are working correctly.

Merged, thanks for the fixes!

PR #1262, opened 2026-08-17
Scale remaining AI turn clocks by game speed

raphaelmonticello:
1. many AI clocks used raw turns as if one turn were one year. on Epic and Marathon natives raid too early, memories expire too soon, and gift attitude is worth less for the same gold.

2. keep counters as raw +1 per turn. convert at the use site with one shared helper:
   - AI_adjustedTurn(n) for "wait N Normal turns" thresholds and contact/memory delays
   - AI_effectiveTurns(raw) for elapsed counters (GrowthPercent)
   - AI_effectiveTurns(raw, percent) for TrainPercent clocks (refuse-to-talk, 10/20/30 peace)
   - AI_speedChance(percent) for per-turn native raid rolls

3. applied to contact delays, memory decay, war/peace/open-borders/DP/share-war attitude, peace-treaty length, native raid start and chance, wander, yield gift/trade timers, scout/founding/city-acquire windows, gift/trade attitude, refuse-to-talk, permanent alliance 40-turn check, and the 10/20/30 peace clocks. the short tactical AtWar check stays on AI_adjustedTurn.

4. greeting / thanks / attitude hover show Total gifts given only after a gift (hidden at 0). refuse-to-talk hover/dialog shows remaining turns until peace talks.

5. native teaching stays in #1258. this PR owns the general game-speed scaling so the two PRs no longer overlap on AI_effectiveTurns.

6. needs a new DLL. no save-format change. the compiled binary is not in this PR.


Schmiddie1979:
Thanks for taking another look at this and expanding the original fix into a more systematic solution. This is much closer to what I had in mind when reviewing the previous PR.

The original issue with the native gift/trade timer was clearly valid, but after looking through the code I found that there are quite a few other AI counters and fixed turn thresholds which potentially suffer from the same general problem. Therefore, I think it is much better to address the underlying issue consistently rather than fixing only one individual counter.

I particularly like the distinction introduced here between:

```cpp
AI_adjustedTurn(...)
```

for fixed turn thresholds,

```cpp
AI_effectiveTurns(...)
```

for interpreting counters which continue to count actual game turns, and

```cpp
AI_speedChance(...)
```

for probabilities which are evaluated once per turn.

Keeping the actual counters as simple raw turn counters is also important. I would prefer not to change the way the counters themselves are incremented. Scaling them when they are evaluated is much cleaner and should also avoid unnecessary savegame-related complications.

I also like that the PR now looks beyond the original native trade/gift counter and covers other areas such as contact delays, memory decay, diplomatic attitude counters, peace treaty timing, native raids/wandering and several AI timing thresholds. This is exactly why I was hesitant to merge the isolated solution from the previous PR before checking the other counters.

There is, however, one issue in `AI_speedChance()` which I think should be fixed before merging:

```cpp
return std::max(1, (iPercent * 100) / iGrowthPercent);
```

A probability of `0` must remain `0`. With the current implementation, a disabled probability could become a 1% per-turn probability.

I would therefore at least add an explicit zero check, for example:

```cpp
if (iPercent <= 0)
{
	return 0;
}

return std::max(1, (iPercent * 100) / iGrowthPercent);
```

There is also a conceptual difference between `AI_adjustedTurn()` and the new `AI_effectiveTurns()`.

The existing `AI_adjustedTurn()` does not use only `GrowthPercent`; it uses the average of Growth, Train and Construct percentages. `AI_effectiveTurns()`, on the other hand, uses `GrowthPercent`.

After checking the standard game speeds, I do not consider this a problem for Normal, Quick, Epic or Marathon because their relevant percentages are aligned. For elapsed-time counters, particularly diplomatic counters, using `GrowthPercent` as the time scale also seems reasonable. I would therefore not try to force all of these cases through one helper merely for the sake of using one formula.

One special case is `GAMESPEED_TAC_BUILDUP_MODE`.

Please **do not add special handling for this game speed to this PR**. Its current game speed values are not properly balanced, and I intend to review and rebalance that game speed separately. The generic implementation should be designed correctly for the normal game-speed system; I will then adjust TAC Buildup's XML values to work with that system rather than adding C++ exceptions for it.

One further point worth keeping in mind is that scaling a per-turn probability linearly is an approximation. For example, halving a per-turn probability while doubling the number of turns does not produce mathematically exactly the same cumulative probability. I do not necessarily consider this a blocker here — particularly for small AI probabilities this may be perfectly adequate — but I think it is important to be aware of the distinction between scaling a fixed turn threshold and scaling a repeated random chance.

Overall, I think this PR is a **significant improvement over the original isolated fix and the general approach is good**. I would like to keep it in review for the moment, mainly to correct the zero-probability case and to make sure that the individual counters are using the appropriate type of scaling. But I think this is now heading towards a proper general solution rather than fixing these counters one by one.

Thanks for taking the broader feedback from the previous PR into account.

One final general note for future contributions:

For changes of this scope, especially changes which affect core gameplay logic, AI behaviour or multiple interconnected systems, I would strongly prefer that the idea is discussed on Discord **before implementation and before opening a PR**.

Changes like this often have consequences beyond the individual issue that initially motivated them. This PR is actually a good example: what started as a fix for one counter turned out to involve a much broader question about game-speed scaling across several AI systems.

Of course, not every small bugfix needs prior discussion. But for more fundamental or far-reaching changes, discussing the concept first makes it much easier to identify related systems, clarify the intended design and agree on the general approach before significant work is invested in the implementation.

So for future changes of this kind, please bring the proposal to Discord first. Once the general approach has been discussed there, the implementation can then be submitted as a PR.

Happy to discuss.

Thank you for your input!

raphaelmonticello:
thanks. I fixed AI_speedChance so a 0% chance stays 0. I did not add anything for TAC Buildup.

I also went through the call sites again:
- fixed turn thresholds, contact delays, memory decay, peace treaty length, raid start, gift/trade timers: AI_adjustedTurn
- elapsed diplomatic counters (has-met, war/peace/open borders/DP/share-war): AI_effectiveTurns with GrowthPercent
- native raid per-turn roll: AI_speedChance

I left the two helpers separate on purpose, same as you said. Linear scaling of a per-turn chance is only an approximation, I know.

for future changes of this size I will bring it up on Discord first.

raphaelmonticello:
1. folded the remaining general game-speed scaling from #1258 into this PR, as Schmiddie asked.
2. one helper: AI_effectiveTurns(raw) for Growth, AI_effectiveTurns(raw, percent) for Train. AI_adjustedTurn and AI_speedChance stay as they were.
3. #1258 is now native teaching only.

PR #1263, opened 2026-08-17
Send rewarded ships to Europe if the capital has no ocean access

raphaelmonticello:
Rewarded ships (king offers, quest rewards, etc.) no longer spawn on a lake or inner sea, and they no longer vanish.

**Where the ship appears**
1. **Capital**, if that city can sail to Europe (`isEuropeAccessable()`: coast that reaches the ocean).
2. Else the preferred city from the event, if it can.
3. Else any other city that can sail to Europe.
4. Else **Europe harbor**, using the same path as buying a ship there (not `initEuropeUnit`, which is the colonist dock and hid the galleon).
5. Else any city, so the unit cannot disappear.

Great river or lake alone is not enough. Closed inner seas fail `isEuropeAccessable()`.

Commandeered Return is unchanged: it still waits for a colonial ocean-access port.

Text/Python only. Restart the game. No DLL rebuild.

### DIFF
--- a/Assets/Python/EntryPoints/CvRandomEventInterface.py
+++ b/Assets/Python/EntryPoints/CvRandomEventInterface.py
@@ -5491,28 +5491,12 @@ def _cityHasEuropeAccess(city):
 	return False
 
 def _cityHasAdjacentOceanPlot(city):
-	# Deep-water berth: ocean or coast. Lake, great river and shallow coast are not enough.
+	# True if the city can sail to Europe: adjacent coast that reaches the ocean.
+	# Closed lakes and inner seas fail this. Great river alone fails this.
 	if city is None or city.isNone():
 		return False
 
-	iOcean = gc.getInfoTypeForString("TERRAIN_OCEAN")
-	iCoast = gc.getInfoTypeForString("TERRAIN_COAST")
-
-	for iDirection in range(DirectionTypes.NUM_DIRECTION_TYPES):
-		pAdjacentPlot = plotDirection(
-			city.getX(),
-			city.getY(),
-			DirectionTypes(iDirection)
-		)
-
-		if pAdjacentPlot is None or pAdjacentPlot.isNone():
-			continue
-
-		eTerrain = pAdjacentPlot.getTerrainType()
-		if eTerrain == iOcean or eTerrain == iCoast:
-			return True
-
-	return False
+	return city.isEuropeAccessable()
 
 def _unitOnCityPlot(player, city, iUnitClass):
 	if player is None or player.isNone() or city is None or city.isNone():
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
Thanks for your fast support. 

In particular, I think introducing one central helper for rewarded ship spawning is a much better solution than maintaining slightly different spawn logic in every individual event. The intended general behaviour should be:

- if the player has a suitable ocean-access city, spawn the rewarded ship there;
- otherwise, spawn the ship in Europe.

This should allow the same logic to be used consistently by Anti-Pirate rewards, Privateer rewards, treasure/ship rewards, trade quest rewards, Cibola-related rewards and the other affected events.

However, after reviewing the proposed implementation, I think there are three points which should be changed before merging.

### 1. Please keep the existing direct `TERRAIN_OCEAN` check

The proposed solution appears to use `isEuropeAccessable()` to determine whether a city is suitable for spawning the rewarded ship.

I would prefer not to use that here.

This problem was investigated previously, and the direct ocean check was introduced deliberately. The important requirement is not merely that the city somehow has access to Europe or to a navigable water area. The city must have a **directly adjacent `TERRAIN_OCEAN` plot**.

The reason is that WTP contains different types of water and ships with different movement restrictions. Lakes and shallow water may be usable by fishing boats, coastal vessels or other special units, while many of the rewarded ships require actual ocean terrain.

Therefore, I would keep the existing helper based on:

```python
def _cityHasAdjacentOceanPlot(city):
	if city is None or city.isNone():
		return False

	iOcean = gc.getInfoTypeForString("TERRAIN_OCEAN")
	if iOcean == -1:
		return False

	for iDirection in range(DirectionTypes.NUM_DIRECTION_TYPES):
		pAdjacentPlot = plotDirection(
			city.getX(),
			city.getY(),
			DirectionTypes(iDirection)
		)

		if pAdjacentPlot is None or pAdjacentPlot.isNone():
			continue

		if pAdjacentPlot.getTerrainType() == iOcean:
			return True

	return False
```

and let the common rewarded-ship helper use this check.

This is important because I explicitly want to prevent ocean-going reward ships from being spawned in inland lakes or in shallow-water-only locations.

### 2. `_spawnRewardedShip()` should return the spawned unit in both cases

The proposed helper currently appears to return the unit when it is created in Europe, but not when it is spawned in a city.

For example, the city path is effectively:

```python
if city is not None and not city.isNone() and _cityHasEuropeAccess(city):
	city.spawnOwnPlayerUnitOnPlotOfCity(iUnitClass)
	return None
```

while the Europe path returns the result of `initEuropeUnit()`.

This creates inconsistent behaviour for callers which need the returned unit.

There are already events which subsequently assign a special name to the rewarded ship, for example ships such as the `Queen Anne's Revenge` or the `Black Pearl`.

If the ship is spawned in Europe, the returned unit is available and the name can be assigned.

If the ship is spawned in a city, however, the helper currently returns `None`, meaning code such as:

```python
unit = _spawnRewardedShip(...)
if unit is not None and not unit.isNone():
	unit.setName(...)
```

will no longer assign the name.

That would introduce a regression.

The helper should therefore return the spawned unit in both cases, for example:

```python
if city is not None and not city.isNone() and _cityHasAdjacentOceanPlot(city):
	return city.spawnOwnPlayerUnitOnPlotOfCity(iUnitClass)
```

and otherwise return the result from `initEuropeUnit()`.

That gives callers one consistent interface regardless of where the ship was created.

### 3. `CommandeeredReturn` should probably remain an exception

After looking at `canDoCommandeeredReturn()` again, I think this particular event should **not** use the general Europe fallback in the same way as the normal rewarded-ship events.

The existing condition:

```python
if getFirstOceanAccessCity(player) is None:
	return False
```

actually makes sense for this event.

Unlike the other cases, `CommandeeredReturn` does not really represent the player receiving a newly awarded ship. It represents the **return of a previously commandeered ship to the colony**.

For normal rewarded ships, I think the intended general behaviour should be:

```text
suitable ocean city available -> spawn in that city
no suitable ocean city        -> spawn in Europe
```

For `CommandeeredReturn`, however, I think the behaviour should remain:

```text
suitable ocean city available -> ship can return there
no suitable ocean city        -> return event cannot occur yet
```

Having a previously commandeered ship suddenly appear in Europe simply because the player currently has no suitable colonial port does not seem appropriate for the narrative or mechanics of this particular event.

Therefore, I would keep the ocean-city requirement in `canDoCommandeeredReturn()` and treat this event as an explicit exception to the general rewarded-ship fallback.

I would also consider keeping the actual spawn logic explicit in `applyCommandeeredReturn()` rather than routing it through a helper which automatically falls back to Europe. This avoids creating a hidden dependency on the `canDo` check: if somebody changes `canDoCommandeeredReturn()` later, the ship should not accidentally start spawning in Europe.

Something along these lines would make the intended behaviour very clear:

```python
city = getFirstOceanAccessCity(player)

if city is None or city.isNone():
	return

unit = city.spawnOwnPlayerUnitOnPlotOfCity(iUnitClass)
```

### Overall

I strongly support the general idea of this PR.

Centralizing the spawning of **actual rewarded ships** is exactly the kind of cleanup I would like to have here, and it should prevent this problem from reappearing independently in multiple events.

Before merging, however, I would ask for the following adjustments:

1. Keep the direct adjacent `TERRAIN_OCEAN` requirement instead of replacing it with `isEuropeAccessable()`.
2. Make `_spawnRewardedShip()` return the spawned unit for both city and Europe spawns, so callers such as the `Queen Anne's Revenge` and `Black Pearl` events can reliably modify the created unit.
3. Treat `CommandeeredReturn` as an exception to the general Europe fallback. Keep its ocean-city requirement and preferably keep its city-only spawn logic explicit.

With those changes, I think this could become a very good central solution for the known rewarded-ship spawning problem while still preserving the special logic of events where a Europe fallback would not make sense.

Schmiddie1979:
Could you please check your version of CvRandomEventInterface.py? Version = current develop?

raphaelmonticello:
Thanks, implemented the three points:

1. Rewarded ships now require a directly adjacent TERRAIN_OCEAN plot (`_cityHasAdjacentOceanPlot`). `isEuropeAccessable()` is no longer used for this.

2. `_spawnRewardedShip()` returns the created unit in both paths. `city.spawnOwnPlayerUnitOnPlotOfCity()` is void in the current DLL, so the city path uses `player.initUnit(...)` (same profession / plot as the city helper) so Queen Anne's Revenge and Black Pearl can still be named.

3. `CommandeeredReturn` is an exception again: `canDo` still requires an ocean city, and `applyCommandeeredReturn()` spawns only on that city with no Europe fallback.

raphaelmonticello:
yes. Version = current develop

Schmiddie1979:
### General point: the affected events need to be reviewed individually

Before going into the individual implementation details, I think there is one important general point about this PR.

I do not think this issue can safely be solved by applying one general replacement to every event which happens to spawn a ship.

The central `_spawnRewardedShip()` helper itself is a good idea, but it seems that **each affected event needs to be reviewed individually before deciding whether it should use that helper and whether the Europe fallback is appropriate for that particular event**.

The event code contains different types of ship spawns which may look technically similar, but do not necessarily represent the same gameplay situation.

For example:

```text
normal reward ship
-> colonial ocean port if available
-> otherwise Europe
```

is a perfectly reasonable general rule.

But `CommandeeredReturn` already shows why this cannot simply be applied to every matching piece of code:

```text
commandeered ship returning
-> requires a suitable colonial port
-> otherwise the return event should wait
```

This is not merely a technical distinction. It is part of the logic and narrative of the individual event.

There may be other cases with similar special requirements. Some events may depend on a particular triggering city, some may require access to the newly created unit afterwards, and some may have `canDo...` conditions which are directly related to where or when the ship is supposed to appear.

Therefore, before replacing all of these spawn calls with `_spawnRewardedShip()`, I would like the affected events to be checked **event by event**:

1. What exactly is the ship representing in this event?
2. Is spawning it in Europe an appropriate fallback?
3. Should it spawn in the triggering city, another suitable city, or specifically the first ocean-access city?
4. Does the event have a corresponding `canDo...` condition which depends on the spawn location?
5. Does the caller need the returned `CyUnit` afterwards, for example for naming or other modifications?

Only after answering these questions should the individual event be moved to the common helper.

So I would prefer not to treat this PR as a general search-and-replace refactoring. The helper can provide the common technical implementation, but the decision to use it still needs to be made separately for every affected event.

This is particularly important in `CvRandomEventInterface.py`, because many events contain special-case logic which is not obvious from the individual spawn call alone.

**One final note from my side: sorry, I initially thought this would be a relatively straightforward fix.**

I did not sufficiently consider that the individual events are structured quite differently and that understanding the context and intended behaviour of each event is important before changing its spawn logic. `CommandeeredReturn` is already a good example of this.

Fortunately, this does not mean that the entire event system needs to be reviewed. I already have a fairly good idea which events are affected, and I expect that we are talking about perhaps 10–20 events rather than all ~300 events in the file.

Still, I do not want to ask you to work through all of those event chains and first become familiar with their individual design just for this PR. I know the event structure and the intended behaviour of many of these events better, so it probably makes more sense for me to take a closer look at the affected cases and determine which ones can safely use the common helper and which ones require their existing special handling.

So please do not interpret my comments as a request to review the entire event system yourself. The central helper is still a useful idea and I think it can become part of the solution. I simply underestimated how much event-specific knowledge is required to decide where the general fallback can safely be applied.

raphaelmonticello:
the helper now follows your rule: any city with adjacent TERRAIN_OCEAN first, Europe only if none exists. It still returns the unit. CommandeeredReturn stays a city-only exception. I also stopped intercepting the generic spawnOwnPlayerUnitOnSamePlotAsCity, so XML events are unchanged.

the apply*RewardShip callbacks still use the helper. if any of those should not fall back to Europe, say which and I will take them out.

Schmiddie1979:
great, thanks, will have a look at it.

raphaelmonticello:
I'm making more changes on it as I'm testing it.

Schmiddie1979:
Thx, I will take it from here.

Schmiddie1979:
Thanks for your support. I have done a fix by myself, cause I created all the events and knew some further critical issues that needed to be solved and which came up during review.

PR #1264, opened 2026-08-18
Show the clipped slave-trader quest reward in the event popup

raphaelmonticello:
The slave-trader quest popup was clipping the last line of TXT_KEY_EVENTTRIGGER_SLAVE_TRADER_QUEST. Players only saw "If we are the first they promise to pay fo experienced" and never **slave overseers**.

The original flavour text is unchanged. The only edit is a [NEWLINE] before the last sentence so the reward wraps onto a new line and stays visible.

Text-only. Restart the game to load it. No DLL rebuild.


Schmiddie1979:
Thanks for the fix.

Keeping the original text unchanged and only adding the line break is the right solution here. This is a small and well-contained text/UI fix with no impact on gameplay logic.

I have therefore accepted the PR.

PR #1265, opened 2026-08-18
Ask how much cargo to offer in native diplomacy

raphaelmonticello:
Clicking a yield on the tribe trade screen dumped the whole stack onto the table. Gold already asks how much. This adds that for cargo, but only on **shift-click**.

- Normal click: whole stack, same as before.
- Shift-click: type how many to offer. Only that many goes on the table and is unloaded if the deal is accepted.
- Leftover cargo stays in Tradeable (900 left after offering 300 of 1200).
- Taking the yield off the table restores the full stack.
- AI still offers the full amount.

The amount box is the existing `Enter amount of X to transfer` spinbox, shown as a Python popup. A DLL popup sits behind the diplomacy screen, so that path does not work here.

No savegame change. The chosen amount is unsaved session state on the player.

Playtested: shift-click 300 of 1200 Trade Goods, leftover 900 stays listed, clearing the offer restores 1200.
image


Schmiddie1979:
This will definitely be a quality of life feature that will be highly appreciated by the community. I will have a look at it asap.

Schmiddie1979:
I hope I will find the time tomorrow to review this feature.

raphaelmonticello:
Hi,

1. that AI_adjustedTurn() on PEACE_TREATY_LENGTH is not needed for the cargo amount selection
2. it came from another change I was testing (scaling turn timers with game speed) and leaked into this PR because both touch CvDeal.cpp
3. the cargo feature only needs the part that loads/unloads the chosen amount instead of the whole stack

I reverted those three PEACE_TREATY_LENGTH lines so this PR stays only about choosing how much cargo to offer natives.

raphaelmonticello:
Hi,

1. the loading side was intentional. Shift-click works on both columns of the trade screen, so the same chosen amount is used when you give cargo from the ship (unload into the village) and when you take cargo from the village (load onto the ship). If no amount was chosen, it still transfers the whole stack, same as before. If you prefer this only for cargo offered to natives, I can restrict it to the unload path.

2. you were right about cleanup. Completing a deal already resets that yield, and taking the yield off the table already clears it. Cancelling diplomacy did not. I added a clear when a new diplomacy session starts and when the deal is cancelled, so an old amount cannot leak into a later trade.

raphaelmonticello:
@Schmiddie1979 I just finished testing it. all good

raphaelmonticello:
1. found the bug: partial sales used unloadStoredAmount, which calls canUnload(). that is false in native villages because they are not on your team, so gold was paid and the cargo stayed
2. full stack sales already used setTransportUnit(NULL), which dumps into the village without that check. that is why only the new amount-selection path broke
3. restored the feature plus this fix in https://github.com/We-the-People-civ4col-mod/Mod/pull/1272

PR #1266, opened 2026-08-18
Stop Python log crashes on city names with accents

raphaelmonticello:
`onCityGrowth` logs the city name through `CvUtil.pyPrint`. Python 2.4 then writes that string to an ascii stdout. If the name has a character like e-acute (Quebec, etc.), the player gets a stack of `UnicodeEncodeError` popups. Alt-tabbing back into the game is when the queued growth event often fires, so it looks like a focus bug. It is just the log line.

This encodes debug output as ascii with `?` for non-ascii letters. The city still grows. Gameplay is unchanged.

Python only. Restart the game to pick it up. No DLL rebuild.

### DIFF
--- a/Assets/Python/CvUtil.py
+++ b/Assets/Python/CvUtil.py
@@ -152,13 +152,22 @@ def myExceptHook(type, value, tb):
 	#total = pre+mid+post
 	total=mid
 	if SHOWEXCEPTIONS:
-		sys.stderr.write(total)
+		sys.stderr.write(_toAsciiLog(total))
 	else:
-		sys.stdout.write(total)
+		sys.stdout.write(_toAsciiLog(total))
+
+def _toAsciiLog(stuff):
+	if stuff is None:
+		return ""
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
Are you sure that that caused the issue? I wonder why no player has reported such bug for such a long time...

RaphaelDeMonticello:
Yup. The error was clear.
Btw, it happened when I alt-tab out and in from the game. So it may be a very specific situation to reproduce it, but it indeed happened and the error was clear.

The fix avoid it to happen again

Schmiddie1979:
Tested locally with a non-ASCII city name (`São Salvador`).

City growth is now logged as:

`PY:S?o Salvador has grown`

without triggering a Unicode encoding exception, and logging continues normally afterwards.

I also checked the regular `pyPrint()` output and did not observe duplicated `PY:` prefixes or additional blank lines.

Works as intended. Thanks!

Issue #877, opened 2023-07-20
Buccaneer Attacks Rival Ships When Moving Onto Colocated Transport


cammcken:
A buccaneer starts on a land tile. Adjacent is a friendly ship and a rival ship on a coastal tile. When trying to move the buccaneer onto the transport, the buccaneer instead engages in combat with the rival ship. (And loses horribly.)

Was able to reproduce using World Builder.

devolution79:
Confirming this as a bug! The core issue is that we are not correctly handling human controlled units that have the bAlwaysHostile property set to 1.

PR #1267, opened 2026-08-19
Let the owner's buccaneer board their own ships

raphaelmonticello:
A buccaneer is hidden nationality for **other players**, not for its owner. After #877 the unit could not board a normal Caravel (or any non-HN transport), so you could not move your own buccaneer on your own ship.

## Change
- `CvUnit::canLoadUnit`: drop the blanket ban on HN land units boarding a non-HN transport. Same-team is already required, so this is the owner's fleet only.
- `CvUnit::canMoveInto`: if the move is a load onto our ship, do not treat a rival on the same coastal plot as a forced attack. That was the original #877 case (buccaneer + friendly transport + rival ship). Boarding wins; a plot with only enemies still attacks.

## Test plan
- [ ] Buccaneer loads onto the owner's Caravel / merchantman (same team, not a pirate hull).
- [ ] Buccaneer still cannot board a **foreign** ship.
- [ ] Coastal tile with **our** transport and a **rival** ship: moving the buccaneer onto it boards, it does not suicide-attack the rival.
- [ ] Coastal tile with **only** a rival ship: buccaneer still attacks as before.

## Notes
- DLL change. Rebuilt FinalRelease locally. Binary is gitignored and is not in this PR.
- Isolated from teaching / gift work.

### DIFF
--- a/Project Files/DLLSources/CvUnit.cpp
+++ b/Project Files/DLLSources/CvUnit.cpp
@@ -3919,14 +3919,20 @@ bool CvUnit::canMoveInto(CvPlot const& kPlot, bool bAttack, bool bDeclareWar, bo
 	{
 		if (bAttack || !canCoexistWithEnemyUnit(NO_TEAM))
 		{
-			if (!isHuman() || (kPlot.isVisible(getTeam(), false)))
+			// Prefer boarding our own ship over attacking a rival on the
+			// same plot (issue #877).
+			const bool bBoarding = !bAttack && !bIgnoreLoad && canLoad(&kPlot, false);
+			if (!bBoarding)
 			{
-				if (kPlot.isVisibleEnemyUnit(this) != bAttack)
+				if (!isHuman() || (kPlot.isVisible(getTeam(), false)))
 				{
-					//FAssertMsg(isHuman() || (!bDeclareWar || (pPlot->isVisibleOtherUnit(getOwnerINLINE()) != bAttack)), "hopefully not an issue, but tracking how often this is the case when we dont want to really declare war");
-					if (!bDeclareWar || (kPlot.isVisibleOtherUnit(getOwnerINLINE()) != bAttack && !(bAttack && kPlot.getPlotCity() && !isNoCityCapture())))
+					if (kPlot.isVisibleEnemyUnit(this) != bAttack)
 					{
-						return false;
+						//FAssertMsg(isHuman() || (!bDeclareWar || (pPlot->isVisibleOtherUnit(getOwnerINLINE()) != bAttack)), "hopefully not an issue, but tracking how often this is the case when we dont want to really declare war");
+						if (!bDeclareWar || (kPlot.isVisibleOtherUnit(getOwnerINLINE()) != bAttack && !(bAttack && kPlot.getPlotCity() && !isNoCityCapture())))
+						{
+							return false;
+						}
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
Thanks for the PR! Your approach pointed me in the right direction and helped me identify the actual interaction between boarding and attacking.

I agree with the main goal of the PR: **Hidden Nationality units should be able to use friendly transports just like normal units.** Hidden Nationality should not by itself prevent a land unit from boarding a friendly non-HN ship.

However, after looking into the original reason for the restriction (#877) and testing the different cases, I decided to solve it slightly differently.

The existing restriction in `CvUnit::canLoadUnit()` was essentially a workaround for #877:

```cpp
// Prohibit a unit with hidden nationality from boarding a non-hidden nat. transport
// This also fixes the bug that allows buccanneers to attack a ship if a friendly ship is
// present on the same plot (see #877)
if (m_pUnitInfo->getDefaultUnitAIType() != UNITAI_YIELD &&
	m_pUnitInfo->isHiddenNationality() &&
	!pTransport->getUnitInfo().isHiddenNationality())
{
	return false;
}
```

I removed this restriction completely. `canLoadUnit()` already checks that the transport belongs to the same team, so there is no need for an additional Hidden Nationality restriction here.

This exposed the original #877 issue again: a land unit normally cannot enter a water plot, but `canLoad()` provides an exception if a suitable friendly transport is present. If an enemy ship is present on the same plot, this boarding exception could also result in the move being interpreted as an attack.

I therefore separated the two concepts: **being allowed to board a transport must not imply being allowed to attack a unit on that water plot.**

First, in `CvUnit::canMoveInto()` the loading exception is no longer allowed to make an otherwise inaccessible water plot valid for an attack:

```cpp
// old
if (bIgnoreLoad || plot()->isWater() || !canLoad(&kPlot, false))

// new
if (bIgnoreLoad || bAttack || plot()->isWater() || !canLoad(&kPlot, false))
```

The corresponding Large River loading checks were adjusted in the same way:

```cpp
// old
bLandUnitMayBeLoaded = canLoad(&kPlot, false);

// new
bLandUnitMayBeLoaded = !bAttack && canLoad(&kPlot, false);
```

and:

```cpp
// old
if (bIgnoreLoad || !canLoad(&kPlot, true))

// new
if (bIgnoreLoad || bAttack || !canLoad(&kPlot, true))
```

There was one additional complication. `MISSION_MOVE_TO` tries `groupAttack()` before proceeding with the normal movement/boarding logic. Therefore, when the destination is a water plot and a land unit can board a friendly transport there, I skip this attack attempt and continue with the normal movement logic instead.

Finally, a normal move into a plot containing an enemy unit would still be rejected by the visible-enemy check in `canMoveInto()`. For an actual boarding move I therefore added:

```cpp
const bool bBoarding = !bAttack && !bIgnoreLoad && canLoad(&kPlot, false);
```

and changed:

```cpp
if (kPlot.isVisibleEnemyUnit(this) != bAttack)
```

to:

```cpp
if (!bBoarding && kPlot.isVisibleEnemyUnit(this) != bAttack)
```

This allows the unit to board its friendly transport even if an enemy ship happens to occupy the same plot, without turning that boarding permission into an attack permission.

I tested the relevant cases and the resulting behavior is now:

- HN land unit + friendly normal transport → **boarding works**
- HN land unit + enemy ship only → **no attack through the loading exception**
- HN land unit + friendly transport + enemy ship on the same plot → **boards the friendly transport instead of attacking**

So the final solution is somewhat different from the implementation proposed here, but your PR was a very useful starting point. In particular, the idea of distinguishing an actual boarding move from the normal enemy-unit handling was exactly the direction needed to replace the old HN restriction with a proper fix for #877.

Thanks for bringing this up and for the proposed solution!

PR #1268, opened 2026-08-19
Stop Inland Seas from consuming large flat areas

StreamlineDeet:
It seems to be a common complaint that inland seas are far too common with the current map script. When investigating a way to reduce the number created, I discovered the cause. The function that generates lakes is designed so that you give it a local minimum, and it fills up the basin that minimum is in until it can overflow and flow to the ocean. However, it treats large flat areas as basins, which is what causes the inland seas. They are simply large flat areas that have a single low point, and the lake in that low point consumes the entire region, submerging it. 
It appears to have been a simple case of the wrong comparison operator, so this PR fixes that.

As a side benefit, this restores large plains to the map script, as they were previously extremely likely to be consumed by inland seas.

### DIFF
--- a/PrivateMaps/FaireWeatherTweakEx.py
+++ b/PrivateMaps/FaireWeatherTweakEx.py
@@ -3883,7 +3883,7 @@ class RiverMap :
             (height,x,y,idx)=k.getLowestBorder()
             ii = GetIndex(x,y)
             
-            if height < k.waterH:
+            if height <= k.waterH:
                 # drainage pathway found! Lake finished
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
Thanks for finding this.

The change makes sense to me. A border plot at exactly the current water height should already provide a drainage path, so requiring it to be strictly lower can indeed cause large flat regions to be treated as enclosed basins and flooded unnecessarily.

Since the change only affects the equality case, it should leave actual depressions unchanged while preventing the oversized inland seas you described.

PR accepted. However, we should review what happens and if the script works as intended.

PR #1272, opened 2026-08-21
Restore cargo amount selection for native diplomacy

raphaelmonticello:
Restores #1265 after it was reverted in #1271.

Shift-click cargo amount selection is back. The bug that paid gold without taking the goods is fixed.

**Bug**
Selling a partial stack (shift-click 50 of 100) paid the gold but left the cargo on the wagon/ship. Full-stack sales still worked.

Partial sales used `unloadStoredAmount()`, which calls `canUnload()`. That check is false in native villages because they are not on your team. Full stacks already used `setTransportUnit(NULL)`, which dumps into the village without that check.

**Fix**
Partial native sales now dump the chosen amount into the village city the same way a full stack does. Leftover cargo stays on the wagon/ship.

Normal click is still the whole stack. Shift-click still asks how much. No savegame change. DLL needed.

Playtest: shift-click sell 50 of 100 to a native village, gold is paid, wagon has 50 left.


raphaelmonticello:
1. this is #1265 restored after the revert in #1271
2. the extra commit is the unload fix: partial native sales now dump the chosen amount into the village the same way a full stack already did
3. please recompile. shift-click sell 50 of 100, gold is paid, 50 should leave the wagon

Schmiddie1979:
Thank you. Tested and works perfect. I merged it.

PR #1273, opened 2026-08-22
Jump to native villages from the domestic advisor row icon

raphaelmonticello:
On the Domestic Advisor native tab, hovering a village icon updated a leftover "Jump to Pawnee" header and did not zoom. City rows already jump when you click the square on the left.

This makes native rows work the same way: click the village icon to look at that settlement. Hover help names that village. The extra header button is removed.

Python only. Restart the game.


RaphaelDeMonticello:
Image of the bug (before the fix)
image

Schmiddie1979:
Thank you for the fix. Works as intended in my test game.

PR #1274, opened 2026-08-22
Fix Shift-F2 Africa, chipotle city peek, and mouse-wheel to close the colony screen

raphaelmonticello:
1. Shift-F2 is already Africa in `CIV4ControlInfos.xml`. With chipotle on, Python was stealing that key to select the city under the mouse, so Africa never opened. Shift-F2 now opens Africa, Ctrl-F2 Port Royal, F2 Europe.

2. Chipotle city peek is kept, just not on Africa's key. Double-click an AI/other-player city on the map, or Ctrl-Shift-F2 on the city under the cursor. Own cities still open on a normal click.

3. Mouse-wheel click closes the colony screen from anywhere on it. The EXE never sends a middle-button event to Python, so the DLL reads the button each frame.

4. Restart the game so both the DLL and Python load.

### DIFF
--- a/Assets/Python/CvEventManager.py
+++ b/Assets/Python/CvEventManager.py
@@ -273,6 +273,17 @@ class CvEventManager:
 				CvScreensInterface.showAchieveAdvisorScreen()
 # Achievements END
 
+			# F2 Europe is the EXE control. Shift-F2 / Ctrl-F2 used to be eaten by
+			# the chipotle city-select cheat, so Africa never opened. Handle those
+			# chords here the same way F10 opens Achievements.
+			if (theKey == int(InputTypes.KB_F2)):
+				if self.bShift and not self.bCtrl:
+					CvScreensInterface.showAfricaScreen((-1,))
+					return 1
+				elif self.bCtrl and not self.bShift:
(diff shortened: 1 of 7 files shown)


raphaelmonticello:
1. nightinggale is right that the old Shift-F2 cheat peeked at AI cities. I kept that, just not on Africa's key

2. double-click does reach Python, but in onMouseEvent, not the keyboard handler. with chipotle on, double-click a city on the map now peeks at it, including AI cities. Ctrl-Shift-click stays Place Object

3. Ctrl-Shift-F2 also peeks at the city under the cursor when cheats are on, as nightinggale suggested. Shift-F2 is still Africa, Ctrl-F2 is still Port Royal, F2 is still Europe

4. Tish: middle mouse (wheel click) now closes the colony screen if the EXE reports that button on a widget click. the camera flying-mouse control is Alt-Shift-F, so it should not fight this. python only, restart the game

Schmiddie1979:
Tested all changes successfully:

Shift+F2 opens Africa correctly
Chipotle city peek works with the new controls
Middle mouse button reliably closes the colony screen

Everything works as intended. Merged, thanks!

PR #1276, opened 2026-08-22
Pad ocean on tall maps so globe view does not go black

raphaelmonticello:
1. zooming far out on tall maps (Americas Gigantic 136x256) turns globe view black. this is an EXE texture/lighting limit, not a bad map file. Don_Drochilla's workaround of padding ocean to y/x ~ 1.45 gives the full-map zoom. this PR does that padding in the DLL so nobody has to edit the map.

2. extra ocean is split on both coasts: half west, half east (and south/north on wide maps). east-only pad made east-coast starts sail farther to Europe than west-coast starts. Americas Gigantic 136x256 becomes 177x256 with 20 west + 21 east ocean columns.

3. unpadded saves are shifted, not left with x,y pointing at the new west ocean. plot data is read into the interior, then cities, units, starting plots, move-to missions, AI mission/attack plots, event plots, and extra-yield plots get the same west/south offset. plot-index caches (AI distance map, best yield plots) are remapped or rebuilt. already-padded 177-wide saves are left alone (ratio is already safe).

4. WB scenarios get the same offset after rebuild, so file coords land on the interior and both new edges copy Europe from the old edges.

5. GLOBE_SPOTLIGHT_ANGLE stays at 30 on huge world sizes instead of interpolating to 90, so lighting does not go edge-on. random gigantic is already 127x184 and is left alone.

6. needs a new DLL. restart the game, then load the save (or start a new game). after you save once, the extra ocean is stored in the save.

7. playtested on an Americas Gigantic save from develop. max zoom-out shows the whole map, no black globe.

### DIFF
--- a/Assets/Python/pyWB/CvWBDesc.py
+++ b/Assets/Python/pyWB/CvWBDesc.py
@@ -1429,6 +1429,31 @@ class CvWBDesc:
 		print("WBSave done\n")
 		return 0	# success
 
+	def applyGlobeviewPadOffset(self):
+		iWest = CyMap().getGlobeviewPadWest()
+		iSouth = CyMap().getGlobeviewPadSouth()
+		if iWest == 0 and iSouth == 0:
+			return
+		if self.plotDesc != None:
+			for pDesc in self.plotDesc:
+				pDesc.iX = pDesc.iX + iWest
+				pDesc.iY = pDesc.iY + iSouth
+				if pDesc.cityDesc != None:
+					pDesc.cityDesc.plotX = pDesc.cityDesc.plotX + iWest
+					pDesc.cityDesc.plotY = pDesc.cityDesc.plotY + iSouth
+				for u in pDesc.unitDescs:
+					u.plotX = u.plotX + iWest
+					u.plotY = u.plotY + iSouth
+		if self.signDesc != None:
+			for pDesc in self.signDesc:
+				pDesc.iPlotX = pDesc.iPlotX + iWest
+				pDesc.iPlotY = pDesc.iPlotY + iSouth
+		if self.playersDesc != None:
+			for pWBPlayer in self.playersDesc:
+				if pWBPlayer != None and pWBPlayer.iStartingX != -1:
+					pWBPlayer.iStartingX = pWBPlayer.iStartingX + iWest
+					pWBPlayer.iStartingY = pWBPlayer.iStartingY + iSouth
+
 	def applyMap(self):
 		"after reading setup the map"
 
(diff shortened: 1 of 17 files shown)


raphaelmonticello:
the solution as it is is working, but it may be a little unfair to east civilizations to get the ocean increased only on their side. I'm evaluating to add on west as well (but it has it's own challenges as it needs to actually pad the map)

raphaelmonticello:
ok, it's actually working well!
When an non-adapted map is loaded, during loading, the game will first adapt it adding ocean to the west and east.
on first turn, those ocean tiles will have no weather/clouds, but when the turn is passed, the game will populate them properly.

raphaelmonticello:
this one needs broader test on multiple maps and loading old saves.
the goal is to not break anything and make everyone happy with the blackness bug fix.
I'll need more people to test it

raphaelmonticello:
follow-up from a We-The-People Americas Gigantic save (Tish) pointed at develop: the EXE aborted on load (VC runtime, process closed).

cause: globe pad-on-load ran setTerrainType(OCEAN) on the extra west/east columns *during* Save_Plots. in the save, Plots are written *before* Areas. those pad tiles also got init() a second time (uninit + gDLL destroy).

this push:
- still pads unpadded gigantic saves on load (136x256 -> 177x256, west+east ocean)
- applies ocean on the pad *after* the map (including Areas) is fully read
- does not init pad plots twice
- skips null areas in the bonus/improvement cache (pad tiles have INVALID area until pad-areas runs)

needs the new DLL. restart, load Tish on WeThePeople-develop. globe should still get the extra ocean; first turn may still have empty weather on the new tiles.

raphaelmonticello:
Tish save still aborted after the last push. growing the plot array during load (136x256 -> 177x256) is what kills We-The-People Americas Gigantic saves, not only setTerrainType timing.

this push stops resizing on save load. plot read is the same linear loop as 4.2.1. globe pad still runs on new games and WB. already-padded 177-wide saves load as-is.

 on develop: restart, new DLL, load Tish. globe may go black on this old gigantic save; the game should stay up. new gigantic games still get west+east ocean.

raphaelmonticello:
Reviewed this after the Tish load tests.

Pad only on **new games / WorldBuilder** (`CvMap::init`). We do **not** grow the plot array while reading a save. Old saves keep their original map size (Tish / Americas Gigantic stays 136x256), so they can still get the black globe. Saves that were already padded (new games started after this) load as-is.

Growing the map in the middle of `Save_Plots` aborted the EXE. A post-load remap (pad after the file is fully read, then shift city/unit coords) would be a later follow-up if we want old saves to get the globe fix too.

Taking this out of draft.

PR #1277, opened 2026-08-22
Stop native braves from merging into one roaming doomstack

raphaelmonticello:
1. native braves were stacking 100+ on one tile. after a village is abandoned to cultural pressure, all the homed units sit on that plot and AI_group(UNITAI_DEFENSIVE) had no size or path cap, so the whole tribe became one roaming stack.

2. homeless braves now look for a new village first. they pick the one with fewer defenders so they do not all dump on the nearest town.

3. native warbands cap at NATIVE_DEFENSIVE_GROUP_MAX (5) and only group with units 1 turn away. extras leave an oversized stack on the next AI pass.

4. needs a new DLL. existing saves: the blob breaks up over a few turns. restart the game.

### DIFF
--- a/Assets/XML/GlobalDefinesAlt.xml
+++ b/Assets/XML/GlobalDefinesAlt.xml
@@ -482,6 +482,13 @@
 		<iDefineIntVal>32</iDefineIntVal>
 	</Define>
 	<Define>
+		<!-- 1. native braves used AI_group(UNITAI_DEFENSIVE) with no size or path cap.
+		     after a village was abandoned they all joined one roaming stack. keep
+		     warbands small. -->
+		<DefineName>NATIVE_DEFENSIVE_GROUP_MAX</DefineName>
+		<iDefineIntVal>5</iDefineIntVal>
+	</Define>
+	<Define>
 		<DefineName>BASE_CHANCE_EUROPE_PEACE</DefineName>
 		<iDefineIntVal>500</iDefineIntVal>
 	</Define>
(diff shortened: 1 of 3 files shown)


Schmiddie1979:
I have integrated this into the native-military-and-village-growth branch and made a few additional adjustments while testing it together with the new Native village growth and military mechanics.

The general approach of redistributing Braves after their home village is lost works very well and should prevent the old problem of homeless Native units gradually merging into one huge roaming doomstack.

I made the following adjustments:

Increased NATIVE_DEFENSIVE_GROUP_MAX from 5 to 10. With the improved Native village growth and the resulting stronger tribes, 10 seems to provide a better balance: Native warbands can still become a serious threat, while uncontrolled doomstacks should be prevented.
Raiding Parties (UNITAI_STATE_RAIDING_PARTY) are excluded from this group-size restriction. They already have their own dedicated AI behavior and should remain able to operate as proper raiding forces rather than being split by the normal defensive Brave limit.
I removed the kill(true) fallback for homeless Braves. Instead of deleting Braves when no new home village can immediately be found, they now remain on the map in limited groups while the existing Native resettlement logic attempts to found a new village.
The existing logic still ensures that only one Native unit becomes a UNITAI_SETTLER at a time. Once a new village has been founded, the remaining homeless Braves can find it through AI_findNewHomeColony() and adopt it as their new home.

This should make the behavior following the loss of a Native village much more natural, particularly when a village is displaced by cultural pressure: the tribe is dispersed and resettles rather than its surviving Braves simply disappearing.

I have merged the changes into native-military-and-village-growth, where they can now be tested together with the new Native population growth and military balancing changes. Its not yet merged with develop - but will be soon I assume. 

Initial tests look good, but this will definitely need some longer-term testing to see how the redistribution/resettlement behavior develops over many turns and after multiple villages are lost.

Schmiddie1979:
Thank you, I integrated this in the native overhaul I implemented.

PR #1281, opened 2026-08-23
AI city center: grow short indoor inputs, not only the best export crop

raphaelmonticello:
1. the AI city-center tile used to pick only the cargo yield with the best Europe sell price. a town with a blacksmith/weaver/cigar maker could still plant sugar on the center while ore/cotton/tobacco was empty, so indoor workers sat idle.

2. AI_updateNeededYields for building slots wrote the OUTPUT yield (cloth, muskets) instead of the INPUT those slots consume (cotton, ore). needed-yield is now the consumed yield.

3. city-center score is still Europe price * yield. if needed > stored + produced, add (needed - have) * AI_CITY_PLOT_DEFICIT_WEIGHT * yield. default weight 30, XML tunable.

4. humans who already set a preferred yield are left alone. natives unchanged.

5. needs a new DLL. restart the game. playtest an AI city whose center can grow ore/cotton/tobacco and that has the matching indoor building with little stored input.

### DIFF
--- a/Assets/Python/EntryPoints/CvRandomEventInterface.py
+++ b/Assets/Python/EntryPoints/CvRandomEventInterface.py
@@ -28204,12 +28204,7 @@ def applyThreeGalleonsRewardShip(argsList):
 	if iRewardUnitClass == -1:
 		return
 
-	(city, iter) = player.firstCity(True)
-
-	if city is None or city.isNone():
-		return
-
-	city.spawnOwnPlayerUnitOnPlotOfCity(iRewardUnitClass)
+	_spawnRewardedShip(player, iRewardUnitClass)
 
 
 def getHelpThreeGalleonsRewardShip(argsList):
--- a/Assets/Python/Screens/CvDomesticAdvisor.py
+++ b/Assets/Python/Screens/CvDomesticAdvisor.py
@@ -361,11 +361,14 @@ class CvDomesticAdvisor:
 		return False
 
 	def getWaitingProductionNames(self, pCity):
-		# 1. WTP keeps hammers when you switch production. Schmiddie: also
-		#    flag those parked buildings/units when they still need cargo.
+		# WTP keeps hammers when production is switched.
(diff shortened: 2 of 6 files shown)


Schmiddie1979:
I can currently not confirm that it works. the AI cities I prepared and looked into did not use the right ressource (hoever, they also did not use the output building right, I assume there was another code that leads to the use of food (although city produced enough food) instead of the right yield for indoor production (coffee in this test).

raphaelmonticello:
thanks for the test. you are right, the first version was not enough.

the city center always grows food plus one cargo yield. a citizen is not a farmer on that home tile. coffee indoor consumes coffee berries, not roasted coffee. so if the town still showed food and an empty roaster, three other bits of code were still winning:

1. AI_assignCityPlot ran BEFORE AI_updateNeededYields. the deficit bonus used last turn. a city that just got a roaster still had need = 0, so Europe export price (sugar/cocoa) won.

2. building-slot need only counted if AI_getYieldAdvantage(output) == 100. a town with a coffee house that is not the empire's single best coffee city recorded no berry need at all.

3. AI_citizenProfessionValue hard-zeroed the roaster when stored+produced berries were 0. extra food jobs on the ring still scored high even when food was already enough. nobody grew berries, so the building stayed empty. that is the other code you assumed.

what this push changes:

- needed yields are computed first (workforce-invariant), then the center crop is picked, then citizens.
- if THIS city has the indoor building (slots > 0), it needs that input. no more 100-advantage gate.
- if food stored+produced is more than 2x population consumption, extra food plot jobs are cut to 1/4. plot jobs that grow a short indoor input (berries, ore, cotton) are *3. indoor still stays 0 until that input exists, so they grow the crop first, then staff the building.

the center can still only grow what that one tile can grow. if berries are only on the ring, you should now see pickers there and then the roaster, not more farmers.

needs the new DLL. new game (or at least a city that reassigns). same coffee test: enough food, roaster, berries possible on center or ring, little stored berries.

PR #1282, opened 2026-08-23
AI attack stacks prefer a nearby city; assault ratio is XML

raphaelmonticello:
1. colonial UNITAI_OFFENSIVE stacks called AI_pickTargetCity with MAX_INT, so a stack in Virginia could march to Peru. natives already capped path at 12. colonial stacks now try a nearby cap first, then fall back to the whole map if nothing is in range.

2. that nearby cap is path TURNS, not tiles. XML AI_STACK_ATTACK_NEAR_PATH_TURNS (default 12) is the Standard-map value. it is scaled by actual map height / Standard height (64), then clamped 8-24. tiny ~8, standard 12, huge ~19, gigantic 24. globe west/east pad does not inflate this because it uses height, not width.

3. the 120% (natives 80%) city-assault power ratio was hardcoded. it is now AI_STACK_ATTACK_RATIO / AI_NATIVE_STACK_ATTACK_RATIO.

4. needs a new DLL. restart the game. playtest a colonial war where a stack has a weak city nearby and a capital far away, including on huge/gigantic.

### DIFF
--- a/Assets/XML/GlobalDefinesAlt.xml
+++ b/Assets/XML/GlobalDefinesAlt.xml
@@ -482,6 +482,19 @@
 		<iDefineIntVal>32</iDefineIntVal>
 	</Define>
 	<Define>
+		<!-- 1. colonial stacks used 120%, natives 80%, hardcoded. nearby path was 12 for natives only. -->
+		<DefineName>AI_STACK_ATTACK_RATIO</DefineName>
+		<iDefineIntVal>120</iDefineIntVal>
+	</Define>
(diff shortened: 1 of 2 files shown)


Schmiddie1979:
Hm, I made some tests but whatever I did, enemy units always attacked/walked against my capital city - also if not in range, enemy stack walked towards my capital and not weak city nearby (within range 19). Seems that the code is not working properly so far (I used a huge map). Seems that more tests needed.

PR #1283, opened 2026-08-23
Stop large attack stacks walking 10 turns to absorb one unit

raphaelmonticello:
1. a bombard-led attack stack called AI_group with iMaxPath=10. a 20-unit army would walk up to 10 turns across the colony to pick up one dragoon. AI_STACK_OF_DOOM_LIMIT (32) only caps merged size, not walk distance.

2. the relative-size path cap has to live in AI_omniGroup because only that function knows the other group's size. iMaxPath=10 is still the ceiling. for land UNITAI_OFFENSIVE joins, a path is rejected if it is longer than 1 + 8 * them / us. a stack of 20 walks 1 tile for 1 unit, and still walks farther to merge with a similarly large group.

3. settler/sea grouping is unchanged (different UNITAI). the empty split-slow-units stub is untouched.

4. needs a new DLL. restart the game. playtest a fat offensive stack with a stray unit 8 tiles away: it should not walk 8 turns to absorb it.

### DIFF
--- a/Project Files/DLLSources/CvUnitAI.cpp
+++ b/Project Files/DLLSources/CvUnitAI.cpp
@@ -3200,7 +3200,8 @@ void CvUnitAI::AI_attackCityMove()
 		if ((bombardRate() > 0) && noDefensiveBonus())
 		{
 			// BBAI Notes: Add this stack lead by bombard unit to stack probably not lead by a bombard unit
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
Hm, I'm sure you are right, but I currently think about how to test that properly. I made some tests **without** this fix and I saw no issue. enemy stack always walks towards its original target in the opposite direction and does not care about the lonely unit 5 tiles away.

PR #1284, opened 2026-08-23
Invasion groups wait for escorts; load from quieter facing ports

raphaelmonticello:
1. BBAI left a comment to wait if the invasion group has no escorts. the code did not wait. a loaded transport group could sail with zero UNITAI_ESCORT_SEA / UNITAI_COMBAT_SEA.

2. if escorts are 0, try grouping an escort or combat ship at path 2. if still 0 and AI_getWaterDanger is > 0, skip this turn. empty water still sails, so they do not wait forever.

3. AI_moveToStagingCity used AI_cityThreat as the base score, so a sieged port looked like the best place to load. assault/massing cities now keep transport/escort inertia, add a small floor for facing water, and divide the most-threatened city by 4.

4. needs a new DLL. restart the game. playtest: loaded transports, no escort, enemy ship in the same water should wait. land stack in ASSAULT_MASSING should prefer a facing quiet port over the city currently being raided.

### DIFF
--- a/Project Files/DLLSources/CvUnitAI.cpp
+++ b/Project Files/DLLSources/CvUnitAI.cpp
@@ -5617,7 +5617,25 @@ void CvUnitAI::AI_assaultSeaMove()
 
 			FAssert(getGroup()->hasCargo());
 
-			//BBAI TODO: Check that group has escorts, otherwise usually wait
+			// 1. do not sail an invasion with zero escorts when the water is contested.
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
Hm, not sure what the issue here is. I tried the playtest, but enemy troop ship always does the same - with my own warship nearby or without: Sails to the nearest city, unloads part of the troops in a native city on the way and brings the rest of its troops to the port. No difference if there is my warship nearby or not.

PR #1285, opened 2026-08-23
Own-land attack stacks fight raiders instead of walking past

raphaelmonticello:
1. BBAI left stack-vs-stack combat commented out. the BTS gate was collateralDamage > 0 (catapults). colonial stacks are dragoons/infantry, so uncommenting that as-is would never fire.

2. own land only, group size at least 3: AI_anyAttack(1, 45, 3) then if not ready to assault a city AI_anyAttack(1, 25, 5). this sits above the generic 60% anyAttack.

3. enemy territory is unchanged. 1-2 unit groups are unchanged. the empty split-slow-units stub is untouched.

4. needs a new DLL. restart the game. playtest a war save: our 6-unit stack on our land next to an enemy raid should attack instead of walking toward a city.

### DIFF
--- a/Project Files/DLLSources/CvUnitAI.cpp
+++ b/Project Files/DLLSources/CvUnitAI.cpp
@@ -3007,24 +3007,23 @@ void CvUnitAI::AI_attackCityMove()
 		return;
 	}
 
-	// BBAI TODO: Stack v stack combat ... definitely want to do in own territory, but what about enemy territory?
-	/*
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
I'm not sure if this really works. I did some tests: At war with Portugal I pisitioned my small stack of 4 colonial militia infantry units in enemy territory right beside enemy city and waited. I positioned an enemy stack of 10 European line infantry also on its own land. What happened: They did neither defend their city nor did they attack me - the infantry stack left its city completely undefended and walked to my capital city 50 turns away...

PR #1286, opened 2026-08-24
Fix native buy-bargain always emptying the gold offer

raphaelmonticello:
## What was wrong

Selling to a tribe (player goods, native gold) bargains as expected. Buying from a tribe (player gold, native goods) always died: the gold on the table went to 0, the deal was rejected, and a later failed bargain roll set TimeNoTrade (the NO_TRADE_BECAUSE_BAD_BARGAIN popup / they will not talk).

## Cause

AI_counterPropose bargaining used inventory->head() as the gold node.

buildTradeTable always inserts TRADE_GOLD with m_iData1 = 0. That is only a placeholder. The live price is on the offer list.

- Sell: the first AI counter writes the native gold amount onto the native inventory node, so a later bargain still saw a real price.
- Buy: the player gold lives on the offer list. Inventory gold stays 0. Bargain multiplied 0, wrote 0 gold into the counter, and replaced the player's offer with nothing.

After a couple of those, tryGetNewBargainPriceBuy fails the 20% roll and applyBargainOutcome sets TimeNoTrade.

## Fix

Look up TRADE_GOLD with m_iData1 > 0 on the offer list first, inventory only as fallback. Copy the node into the counter instead of mutating inventory. Floor the bargained price at 1 so we never write 0 gold over a real offer.

Sell path uses the same lookup so it does not depend on the inventory node still holding the last counter amount.

### DIFF
--- a/Project Files/DLLSources/CvPlayerAI.cpp
+++ b/Project Files/DLLSources/CvPlayerAI.cpp
@@ -3980,6 +3980,37 @@ int CvPlayerAI::AI_militaryHelp(PlayerTypes ePlayer, int& iNumUnits, UnitTypes& 
 	return kPlayer.getEuropeUnitBuyPrice(eUnit) * GC.getDefineINT("KING_BUY_UNIT_PRICE_MODIFIER") / 100;
 }
 
+// 1. bargain used inventory->head() gold. buildTradeTable always writes that
+//    slot as m_iData1=0. sell still worked because the first AI counter writes
+//    the native gold amount onto the native inventory node. buy never does
+//    that: the player gold lives on the offer list. reading 0 rewrote the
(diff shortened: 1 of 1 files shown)


raphaelmonticello:
Playtested this more carefully.

The report (buy bargain always rejected, tribe expels the player, sell works) was the existing fail roll, not a broken gold counter.

`BASE_CHANCE_BARGAIN_BUY_FAIL` is 200 / ~1000 (about 20%), same as sell. On fail the tribe sets `TimeNoTrade` and will not talk for several turns, so you cannot retry that village. One fail looks like "always". Trying other tribes, buy bargain does succeed.

Closing this. The inventory-head gold lookup may still be brittle, but it is not what Tish hit and it should not go in as a fix for that.

PR #1287, opened 2026-08-24
Brazilian Portuguese translation

raphaelmonticello:
Brazilian Portuguese (pt-BR) translation.


raphaelmonticello:
1. civ names: Spanish and Swedish colonies now use title case like the others (Colônias Espanholas / Colônias Suecas)
2. Explorers trait: dropped leftover English. it now reads eventos únicos ao explorar em terra / no mar
3. same cleanup on Louis Jolliet and James Cook strategy text
4. XML only

PR #1289, opened 2026-08-25
Stop African and Native slaves from speaking with chiefs

raphaelmonticello:
African slaves and Native slaves can no longer Speak with Chief. The command is hidden for those units.


Schmiddie1979:
Hm, correct me if I'm wrong, but I tested that yesterday and I assume that african slaves already could not do that. Not sure, if we need a code change for that resp. can change that in XML, but I have not looked into the code so far.

Schmiddie1979:
Thanks, that's a good addition. I tested it right away and added “Prisoner of War” as well. I went ahead and updated my version right away. Thanks!

PR #1290, opened 2026-08-25
Skip the war popup for natives; keep it for Europeans

raphaelmonticello:
Native villages no longer ask "does this mean war?" when a military unit moves onto them. European territory still asks, same as before.

1. natives: no popup, no auto-war. MOVE_TO goes through. declare war from the diplomacy screen, then move in to fight
2. Europeans: the popup is unchanged. Yes declares war and moves. No moves only if the tile can be entered without war
3. cargo ships still skip. troop ships still get the European popup
4. refuse-to-talk still offers Declare War in diplomacy when canDeclareWar is true (and Revolution vs the king)
5. DLL. rebuild the DLL

### DIFF
--- a/Assets/Python/CvDiplomacy.py
+++ b/Assets/Python/CvDiplomacy.py
@@ -54,17 +54,13 @@ class CvDiplomacy:
 
 			# We say hi and begin our peace
 			self.addUserComment("USER_DIPLOCOMMENT_PEACE", -1, -1)
-
-			# if you are on different teams and NOT at war, give the user the option to declare war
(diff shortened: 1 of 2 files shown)


Schmiddie1979:
Hm, I'm not convinced so far. I like it as it is...

raphaelmonticello:
that's why it's just a draft like I mentioned in the ticket. 
if not agreed, we can close it. no problem.

Schmiddie1979:
Hm, at least a tooltipp maybe would make sense to show up with a message player cannot enter foreign territory and should use diplomatic options (he is free to try to open border or declare war..haha).

raphaelmonticello:
adapted this after Schmiddie's QoL ask.

natives: no "does this mean war?" popup when a military unit enters a village. no auto war. if you want to fight, declare war from diplomacy then move in.
europeans: the popup stays as it was.

PR #1291, opened 2026-08-25
Algonquian Braves should not use the musket icon until they have muskets

raphaelmonticello:
Algonquian unarmed Braves (Penobscot etc.) used the musket / helmet portrait from game start.

1. natives atlas cell 1,1 is the vanilla Armed Brave musket icon. RaR assigned all four Algonquian professions to that cell
2. unarmed Brave now uses the vanilla melee club icon (Unit_Resource_Colonization_Atlas 3,4)
3. armed / scout / mounted use the matching vanilla cells so they no longer share the musket portrait
4. XML only. reload the mod

### DIFF
--- a/Assets/XML/Art/CIV4ArtDefines_Unit.xml
+++ b/Assets/XML/Art/CIV4ArtDefines_Unit.xml
@@ -5569,7 +5569,7 @@
 		</UnitArtInfo>
 		<UnitArtInfo>
 			<Type>ART_DEF_UNIT_ALGONQUIAN</Type>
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
For now, I’d appreciate it if the nation had a button that matched the units’ appearance. Either we create a new button now, or let’s just leave it as is for the time being. The one currently in use is meant for enemy Native Americans who don’t belong to a tribe, I think. I’d leave it as is for now. The buttons are only meant to represent the tribe in a general sense anyway and aren’t tailored to every conceivable unit type. When I tackle my feature to overhaul all units, I’ll add that to the list as well. Thanks .

PR #1292, opened 2026-08-25
Reduced Colony Distance for 2-plot radius

raphaelmonticello:
1. Reduced Colony Distance was only dropping the min founding range if you already owned the 3x3 around the new site. with 2-plot city that still blocked founding at 3 plots on empty land.

2. with 2-plot radius the option now always reduces min distance by 1 (closest city is 3 plots). 1-plot radius is unchanged: still needs the owned 3x3.

3. needs a new DLL. the compiled binary is not in this PR. existing saves keep the option, just restart.

### DIFF
--- a/Project Files/DLLSources/CvPlayer.cpp
+++ b/Project Files/DLLSources/CvPlayer.cpp
@@ -6094,7 +6094,8 @@ int CvPlayer::getFoundCityMinRange(const CvPlot* pPlot) const
 
 	// 2-plot radius: always reduce by 1. 1-plot keeps the 3x3 ownership check
 	// so close colonies are for filling gaps, not the first wave.
-	if (GC.getGameINLINE().isOption(GAMEOPTION_TWO_PLOT_CITY_RADIUS))
+	// Use CITY_PLOTS_RADIUS: random maps can set radius without the game option.
+	if (CITY_PLOTS_RADIUS == 2)
 	{
 		return iRange - 1;
 	}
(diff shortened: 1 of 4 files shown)


Project Files/DLLSources/CvPlayer.cpp
```diff
@@ -6084,6 +6084,41 @@ void CvPlayer::doGoody(CvPlot* pPlot, CvUnit* pUnit)
...
+{
+	int iRange = GC.getMIN_CITY_RANGE();
+	if (!GC.getGameINLINE().isOption(GAMEOPTION_REDUCED_CITY_DISTANCE) || iRange <= 1)
+	{
+		return iRange;
+	}
+
+	// 2-plot radius: always reduce by 1. 1-plot keeps the 3x3 ownership check
+	// so close colonies are for filling gaps, not the first wave.
+	if (GC.getGameINLINE().isOption(GAMEOPTION_TWO_PLOT_CITY_RADIUS))
```
Nightinggale:
```suggestion
	if (CITY_PLOTS_RADIUS == 2)
```
Random maps will not set the game option for this, so use the global read only variable for getting colony radius. Colony radius specifically is special because it's a setting in the game, which is set either by map options or a game option. This is a workaround for the main menu interface being in the exe and hence can't be modded.

raphaelmonticello:
ok, switched to CITY_PLOTS_RADIUS. also merged develop (the 1281 city-input fix).

Schmiddie1979:
" also merged develop (the 1281 city-input fix)." means what? Is already merged...

raphaelmonticello:
> lso merged develop (the 1281 city-input fix)." means what? Is already merged...

AI slop doing the comment for me sometimes :)

PR #1293, opened 2026-08-26
No sail-to-Europe popup on the red tiles

raphaelmonticello:
The "Do you wish to sail to Europe?" popup is gone. The ship does not sail on its own. When you want to sail, use the Europe / Africa / Port Royal buttons (or Ctrl+E), same as always.

1. no popup on the red Europe tiles
2. no auto-sail. the ship stays there
3. Africa and Port Royal already had no confirm; unchanged
4. Python only. reload the mod


raphaelmonticello:
The "Do you wish to sail to Europe?" popup is gone. The ship does not sail on its own. When you want to sail, use the Europe / Africa / Port Royal buttons (or Ctrl+E), same as always.

PR #1297, opened 2026-08-29
Offer founding fathers one popup at a time

### LINKED ISSUE #1193: Offering Multiple Founding Fathers In One Turn Despite Insuffiicent Points
There seems to be a somewhat uncommon situation in the latest version where if you have sufficient Political Points, and whichever other points type, it will offer two founding fathers in the same turn, even if the first offered is accepted resulting in the second not being affordable.

In one turn I was offered Peter Minuit (990 Exploration, 866 Political), which I accepted, and was then immediately also offered Pocahontas (825 Political). Since I had accepted Peter Minuit I no longer had 825 Political Points so Pocahontas should not have been offered. If accepted it fails silently (Pocahontas was not acquired).

### LINKED ISSUE #1244: Problem with choosing Founding Fathers
[hans- n. Chr.-1587.zip](https://github.com/user-attachments/files/28933746/hans-.n.Chr.-1587.zip)

**Describe the bug**

In several games, starting around 1550 AD, I encounter a problem where I am no longer asked whether I want to accept or decline a Founding Father.
This issue already existed on my old computer, and I’m experiencing it again on my new one...

See the sreenshots

[hans- n. Chr.-1587.zip](https://github.com/user-attachments/files/28933794/hans-.n.Chr.-1587.zip)

Image

Image

**To Reproduce**
Ideally provide a savegame where this happens consistently, in which case the steps to reproduce would be load, click end turn or something like that.

**Expected behavior**

**Screenshots and savegames**
Savegames often helps a lot when reproducing issues as well as testing fixes, so provide one if a savegame can make it easier to reproduce the problem. Screenshots can be used as well if it helps explaining something on the screen, which isn't obvious with text.

Do note that GitHub does not like the file extensions of savegames, so you need to zip the file first.

**Personal modifications**
Did you change anything in the mod or game, like changing xml data or using world builder?

**Version of We The People**
Version of release or in case of development version, a date for when it was compiled will be fine. Note that this is included in savegames, so if one is provided, the precise development version will not be needed, just a rough estimate.

**Additional context**

raphaelmonticello:
`testFoundingFather` queued a popup for every founding father the human could afford at the start of the turn. Accepting the first one spends the points immediately. The second popup still fired, but `convinceFather` then failed `canConvinceFather` (not enough points left) and returned without recruiting and without setting ignore. The yes click did nothing.

I hit this in a playthrough: two founding fathers were offered back to back, and I accepted both. The first one was recruited, the second wasn't (due the lack of points).

As the player also accepted the second, he thinks the founding father was recruited, but it wasn't. This causes confusion to the player.

This keeps multiple fathers per turn when you still have the points (or if you refuse the first). Humans now get one popup at a time. After yes or no, the next father is offered only if `canConvinceFather` still succeeds.

AI behavior is unchanged: it still takes every father it can pay for, in order, during its turn.

Fixes #1193
Fixes #1244

### DIFF
--- a/Project Files/DLLSources/CvTeam.cpp
+++ b/Project Files/DLLSources/CvTeam.cpp
@@ -1820,9 +1820,12 @@ void CvTeam::convinceFather(FatherTypes eFather, bool bAccept)
 		setFatherIgnore(eFather, true);
 	}
 
-	// This is a recursive call, because we are called from testFoundingFather! Why?
-	// If we disable it, there will only be one FF offered to each team per turn, which is preferable for balancing IMHO:
-	//testFoundingFather();
+	// 1. don't queue the next FF popup until this one is resolved.
+	// 2. after a yes the points are already gone, so canConvinceFather filters out ones we can no longer pay for.
+	if (isHuman())
+	{
+		offerNextFoundingFather();
+	}
 }
 
 bool CvTeam::isHuman() const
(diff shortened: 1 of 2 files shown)


Schmiddie1979:
Hm, interesting. I tried that also in current version without the fix - exactly the same, I saw no difference. Current develop: When I get 2 FF at the same time and accept the first, the second one does not fire - hence, is the fix really needed? Or do I misunderstand something.

Razonatair:
Yeah I thought this had been addressed some time ago.

RaphaelDeMonticello:
I could see the comment in the code from 2024, but the solution was still not enough, so I fixed it

PR #1298, opened 2026-08-29
Show game-speed-scaled waits on pirate event help

raphaelmonticello:
## Problem

On a Marathon playthrough, the harbor pirate event offered to spend 150 muskets for an 80% chance of a pirate ship. The musket cost was already scaled correctly (50 x StoragePercent), but the option help never stated the wait for the ship.

EVENT_PIRATES_4a waits iEventTime 5, which applyEvent scales by GrowthPercent. On Marathon that is 15 turns. The follow-up also has iRevoltTurns 1, which the generic event help scales to 3 turns of disorder. That 3 is the number that showed up as a wait, so the option looked like a 3-turn ship instead of 15.

The same popup's treasure option hardcoded "after 5 turns" in every language, so Marathon still read 5 there as well. Portuguese TXT_KEY_EVENT_DELAY_TURNS still said "in" instead of "em".

## Fix

- getHelpPirates4 now prints the 80% ship chance with the same GrowthPercent delay applyEvent uses.
- getHelpPirates1 / TXT_KEY_EVENT_PIRATES_1_HELP take the scaled wait instead of a hardcoded 5.
- Portuguese delay fragment is "em %d1 turnos".

Python and XML only. Restart the game (or reload the mod) to pick this up. No DLL change.

The 80% roll still happens immediately when the option is accepted. The remaining 20% does not schedule a follow-up and has no fail popup. That part is unchanged.

## Test

- Marathon: muskets option should show 150 muskets and 15 turns for the 80% ship. Treasure option should say 15 turns, not 5.
- Normal speed should still show 5.

### DIFF
--- a/Assets/Python/EntryPoints/CvRandomEventInterface.py
+++ b/Assets/Python/EntryPoints/CvRandomEventInterface.py
@@ -5478,6 +5478,20 @@ def _getPiratesScaledTurns(iBaseTurns):
 	iPercent = gc.getGameSpeedInfo(gameSpeedType).getGrowthPercent()
 	return max(1, int((iBaseTurns * iPercent) / 100))
 
+def _getPiratesFollowupChanceAndDelay(event, szFollowupType):
+	# 1. EventTimes in XML is the standard-speed wait.
+	# 2. applyEvent scales that by growth percent, so the option help has to do the same.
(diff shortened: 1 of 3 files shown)


Schmiddie1979:
Thanks, nice catch.

PR #1299, opened 2026-08-30
Show workshop input yields on city building tooltips

### LINKED ISSUE #1173: Butcher bldg tooltip does not show all produce types
**Describe the bug**
When mousing over the butcher building in city-view, I only see Down and Feathers as reported outputs, even though my butchers are butchering pigs. See screenshot.

**To Reproduce**
1. Produce pigs (or I suspect cows).
2. Mouseover butcher in city-view.

**Expected behavior**
I expect the tooltip to display information relevant to what I'm producing.

**Screenshots and savegames**
Image
Hopefully this link works ^^.

**Personal modifications**
None

**Version of We The People**
Compiled 4.2DEV version, pulled Sept 2, 2025.

**Additional context**
NA

raphaelmonticello:
## Problem

City building tooltips listed output per citizen (for example 3 household goods) but not the inputs needed to make that output. The recipe was only in the civilopedia strategy text. On a village market that meant the hover did not show exotic wood and goat hides next to the goods.

Empty worker slots on a building also passed specialbuilding -1, so that hover showed no building help at all.

Butcher and similar workshops produce two yields. The hover only printed `getYieldsProduced(0)`, so pig butchering looked like downs only (#1173).

## Fix

- Building hover now prints the profession output and the consumed yield icons on the same line, e.g. `3 [goods] per Citizen (Consumes: [exotic wood][goat hides])`.
- Extra produced yields are appended, so butcher shows food plus hides/wool/downs.
- The production-icon hover on the building uses the same consume list.
- Empty worker slots use the building id in data2 so they get the same tooltip.
- Civilopedia building requires list gets the consume icons and extra outputs too.

This applies to every workshop with profession output. Chains without inputs (town hall bells, chapel, school, hospital, court, tavern) keep the old output-only line.

Village market was playtested. No XML change; existing `TXT_KEY_YIELD_CONSUMED` is reused.

Needs a rebuilt DLL.

Fixes #1173


Schmiddie1979:
Thanks, looks gread so far. 

However, I might have found an issue while testing this.

The Butcher profession can produce six different yields, but when hovering over the individual yields in the city interface, only two of them (Down and wool) show the corresponding information in the yield tooltip regarding a sale.

The new building tooltip itself is intended to display all entries from getNumYieldsProduced(), so this seems to be a separate inconsistency in the yield-hover tooltip (CvGameTextMgr::setYieldHelp()) or in the way multi-output professions are accounted for there.

Since the purpose of this PR is also to improve support for professions with multiple outputs, I think this case should be checked before merging. Ideally, all six Butcher outputs should be represented consistently both in the building tooltip and when hovering over the individual yields.

image


input info works good.

raphaelmonticello:
Good catch. The meat building is six butcher professions (sheep, goats, cattle, pigs, chicken, geese), each with two outputs. The city yield icon only wired the first and last primary yields, so the hover showed wool and downs and skipped the rest.

Pushed a follow-up: hovering a yield now also lists every indoor profession this city can run for that yield (recipe + extra produced icons + consume list), not only the ones currently staffed. Needs a rebuilt DLL.

Schmiddie1979:
image
image

still does not work for butcher builkdings regarding prices / demands...production input for worker is fine, but not price overview. Could you please check?

raphaelmonticello:
The worker tooltip was already listing both outputs. The production-icon hover is a different widget: it only stored two yield ids (first butcher profession and last), so you got wool and downs prices and nothing for hides, goat hides, pig skin or food.

Pushed a follow-up. Hovering that icon now expands to every yield the shared indoor professions produce, each with Europe / Africa / Port Royal prices, domestic demand, and the consume list. Needs a rebuilt DLL.

Schmiddie1979:
Oh my god...hehe. All goods are shown, but the list is cut at the end - maybe we should try not to show food, because its an extra yield in order to shorten the overview - or can it be shown side by side, not one below the other?

image

raphaelmonticello:
Agreed, that list was too long. The hover was dumping a full city yield report for every butcher output, including food as a byproduct, so it ran off the screen.

Pushed a shorter version. Hovering the production icon now only lists each profession's *primary* good (hides, wool, downs, goat hides, pig skin) with Europe / Africa / Port Royal prices and domestic demand. Food is skipped here: it is an extra output and already has its own city food icon. Recipes and worker production stay on the worker tooltip, where they fit. Side-by-side columns are not something this tooltip can do. Needs a rebuilt DLL.

Schmiddie1979:
Hm, seems that all changes are gone...and also no output is shown anymore - hence every information is gone  - upload issue?

raphaelmonticello:
Not an upload issue. The PR still has every commit (recipes on the worker tooltip, extra produced icons, then the price hover experiments).

The last compact hover could come out blank, so it looked like everything had vanished. That was a bug in the production-icon hover, not missing files.

Pushed a follow-up: that hover always shows a short price/demand block for each *primary* butcher good (hides, wool, downs, goat hides, pig skin). Food is still skipped there. If that list is empty, it falls back to the original two-yield help so the tooltip is never blank. Worker production details are unchanged.

Please rebuild the DLL from this branch. Python/XML restart is not enough for this one.

Schmiddie1979:
compile error...this seems to be a tough one :-)

image

raphaelmonticello:
Ha, not so tough this time :-)

Assert compile failed on `GC.getNumYieldInfos()`. That exists on the Python/Cy side (`gc.getNumYieldInfos()`), but not on `CvGlobals`. Yields are a compile-time enum, so the C++ bounds check is `NUM_YIELD_TYPES`.

Pushed a one-line fix. Please rebuild the DLL from this branch. The hover itself is unchanged: primary butcher goods (hides, wool, downs, goat hides, pig skin) with prices/demand, food skipped, fallback so it is never blank.

Schmiddie1979:
Hm, looks like this now: 

image

raphaelmonticello:
The five primary goods and the prices are right, and skipping food is right. Chicken and geese both make downs, so five lines is correct, not six.

What is still awkward for the player is the layout. Those Europe / Africa / Port Royal lines were written for the wide city-yield hover. On this production icon they wrap, so Kauf and the buy price sit on two lines and the tooltip becomes a tall strip.

We still cannot do real columns. What we can do is one line per good with all three markets on that row. Example: wool with EU/AF/PR sell/buy together. That is basically the side-by-side idea in this tooltip.

Want me to try that before you merge?

Schmiddie1979:
Sounds good, lets try

raphaelmonticello:
Tried the one-line layout.

Each primary good is now a single row: icon, name, EU/AF/PR sell/buy, gold at the end. Domestic demand only shows on that same row if the city actually has any. Food is still skipped.

Please rebuild the DLL from this branch (and restart so the new text keys load).

PR #1301, opened 2026-08-30
Do not fail untracked quests when a stored unit is gone

### LINKED ISSUE #1068: Giving the king a frigate for a random event can cause an unrelated quest to fail
From the bug reporting channel

> If you give your king a frigate (for a random event) while the quest for obtaining a royal battleship is active, the quest will automatically fail 
becomes even more of a headache if you buy your first frigate fairly late in the game (and if it is your first navy ship), as the event for giving your king that frigate and the quest both trigger, effectively locking you out of the quest 
something similar happens with the slave barque quest: If you have the slave barque quest active and you put a slave in one of your cities to work, the quest for obtaining two slave barques also fails

raphaelmonticello:
`checkExpireEvent` failed any quest whose stored `m_iUnitId` unit was gone.

The six-warships and slave-barque *start* triggers pick a unit only to test "you have N of these". `bTrackUnit` is 0 on those. Giving the king that frigate (commandeered event kills it) or assigning that slave to a city then failed the unrelated quest.

Expire on a missing unit only if the trigger asked to track it.

Needs a rebuilt DLL.

Fixes #1068

### DIFF
--- a/Project Files/DLLSources/CvPlayer.cpp
+++ b/Project Files/DLLSources/CvPlayer.cpp
@@ -16176,7 +16176,12 @@ bool CvPlayer::checkExpireEvent(EventTypes eEvent, const EventTriggeredData& kTr
 		}
 	}
 
-	if (kTriggeredData.m_iUnitId != -1)
+	// Only fail the quest because the stored unit is gone when the trigger
+	// asked to track that unit. Start triggers like six warships / slave
+	// barques pick a unit to test "you have N of these", but bTrackUnit is
+	// 0: giving the king that frigate, or assigning that slave to a city,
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
From my perspective, this is a sensible and logically correct change. A quest should only fail because a stored unit is gone if that unit is actually meant to be tracked. However, with the current code in initTriggeredData(), the unit ID is already only stored when bTrackUnit is enabled. Therefore, the fix may no longer be strictly necessary for newly triggered quests. Nevertheless, it provides an additional safeguard, particularly for existing trigger data or older savegames, and I don't see any harmful side effects.

PR #1302, opened 2026-08-30
Name the planted forest, not the tundra terrain

### LINKED ISSUE #1016: [v4.1 Bug] Tooltip for planting forrest refers to base terrain instead to plantable forrest type
**Version**
The version played is 4.1 .

**What did I do?**
I moved a pioneer on a hills/tundra tile and hoverd over the button for planting a forest.

**Expected result**
The tooltip saying: "Plant northern forrest"

**Observed result**
The tooltip says: "Plant tundra"

v4_1 Bug wrong text on plant forrest button

raphaelmonticello:
The tundra reforestation button said "Plant tundra". It plants `FEATURE_FOREST_TUNDRA` (Northern Forest), not the base terrain.

English / Portuguese / French / German / Russian now name the forest, matching the feature text and the expected "Plant northern forest".

XML text only. Restart the game. No DLL rebuild.

Fixes #1016

### DIFF
--- a/Assets/XML/Text/XML_AUTO_UTF8_BuildInfo.xml
+++ b/Assets/XML/Text/XML_AUTO_UTF8_BuildInfo.xml
@@ -370,11 +370,11 @@
 	</TEXT>
 	<TEXT>
 		<Tag>TXT_KEY_BUILD_REFORESTATION_TUNDRA</Tag>
-		<English>Plant tundra</English>
-		<Portuguese>Plantar tundra</Portuguese>
-		<French>Planter une taïga</French>
-		<German>Taiga aufforsten</German>
-		<Russian>Высадить [COLOR_HIGHLIGHT_TEXT]тундру[COLOR_REVERT]</Russian>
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
I think this question should probably be considered in light of the key "TXT_KEY_BUILD_REFORESTATION_TUNDRA". To me, this seems to refer to the command itself. The question is therefore how the wording should be understood: does "Reforestation Tundra" mean reforesting tundra, i.e. the terrain on which the action is performed, or is the command supposed to describe what is being planted, i.e. the resulting forest type? It would also be useful to compare this with the other reforestation commands: how are they named, and do they refer to the terrain being reforested or to the type of forest being planted?

Schmiddie1979:
Furthermore: I assume the current English translation is wrong - thats right (hence translation mistake), the original German version was probably not and had the meaning of the first alternative

Schmiddie1979:
thx, will add this translation directly in develop version.

PR #1303, opened 2026-08-30
Keep yield cargo on a captured wagon train

### LINKED ISSUE #1181: Capturing a wagon train leaves the yield cargo as standalone units, which vanishes the following turn
Capturing a wagon train leaves the yield cargo as standalone units, which vanishes the following turn.

To reproduce go to La Isabella (far east of the yellow land), select the buccaneer and attack the French wagon trains directly west of it. 4 yield units will be dropped outside of the wagon trains.

Savegame
[yield capture bug.zip](https://github.com/user-attachments/files/22667358/yield.capture.bug.zip)

video
https://cdn.discordapp.com/attachments/1055869225358606356/1423393582500286464/2025-10-02_14-36-56.mp4?ex=68e02613&is=68ded493&hm=3c23436467c7a64cbf6b5b7f44c258424cfdd7f11007d11105ddf809b6a272ac&

raphaelmonticello:
Capturing a wagon train dumped its yield cargo as standalone goods units on the plot. Those vanish the next turn.

Combat no longer captures/kills goods that are loaded on a land wagon. `kill` keeps them on the wagon through delayed death, then loads them onto the captured transport.

Pirate `isCapturesCargo` loot from sunk ships is unchanged.

Needs a rebuilt DLL.

Fixes #1181

### DIFF
--- a/Project Files/DLLSources/CvUnit.cpp
+++ b/Project Files/DLLSources/CvUnit.cpp
@@ -636,12 +636,10 @@ void CvUnit::kill(bool bDelay, CvUnit* pAttacker)
 		pUnitNode = pPlot->nextUnitNode(pUnitNode);
 	}
 
-	// Yield cargo captured as standalone goods units vanishes next turn.
-	// If this transport is itself being captured, remember the yields and
-	// load them onto the new wagon after it is created.
-	std::vector<UnitTypes> aeYieldCargoUnits;
-	std::vector<ProfessionTypes> aeYieldCargoProfessions;
-	std::vector<int> aiYieldCargoStored;
+	// Yield cargo captured as standalone goods vanishes next turn.
+	// Combat leaves it on the wagon. When this transport is captured,
+	// let kill() spawn the goods for the capturing player, then load
+	// those units onto the new wagon.
 	const PlayerTypes eYieldCargoPlayer = getCapturingPlayer();
 
 	for(uint i=0;i<oldUnits.size();i++)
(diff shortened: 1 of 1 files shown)


Schmiddie1979:
Could you please check again. I tried to capture wagons with different goods from enemy but still as in the current develop version only the wagon was captured, not the goods. I double checked whether I had the correct file and recompiled before I tested...

raphaelmonticello:
You were right, the wagon was coming in empty.

The first version destroyed the yield cargo and tried to rebuild it. That rebuild never made it onto the captured wagon.

Pushed a follow-up. Combat still leaves the goods on the wagon. `kill()` now captures those yield units the normal way, then immediately loads the capturer's loose yield units on the plot onto the new wagon so they are not left standing on the ground. Needs a rebuilt DLL.

Schmiddie1979:
Are you sure? Still does not work...no captured goods

Schmiddie1979:
PR #1303 was not merged because the proposed implementation did not reliably preserve captured wagon cargo.

The original version attempted to preserve and recreate the yield cargo during the capture process, but in testing the captured wagon still ended up without its goods.

A later version changed the approach and captured the yield units separately first, then searched the plot for loose captured yield units and loaded them onto the newly captured wagon.

This still did not reliably solve the problem and introduced an additional issue: once the cargo had been separated from its original transport, the original wagon-to-cargo association was lost. When multiple loaded land transports were captured on the same plot, goods could therefore be assigned to the wrong wagon. If that wagon did not have enough cargo capacity, goods could also be lost.

Instead, I implemented a different solution that keeps the original association between each land transport and its cargo intact throughout the capture process. Yield cargo remains attached to its corresponding transport until that transport itself is processed, and is then restored to the newly captured version of that specific transport.

This avoids reconstructing cargo ownership from loose yield units on the plot, handles multiple captured wagon trains correctly, and leaves the existing sea transport capture behavior unchanged.

PR #1304, opened 2026-08-30
Only let a unit that can talk pick the chief

### LINKED ISSUE #1030: Talking to Village Chief bugs out if a unit that cannot talk to chiefs is selected and ahead of the unit that can talk to the chief.
Just experienced this strange little bug. I have a Treasure and a Failed Trader in a Native Village that hasn't been met yet. The Treasure is listed ahead of the Failed Trader. I had them both selected, then clicked the button to talk to the chief, but nothing happened. After that, I cannot talk to the village anymore. I suspect there is a bool getting flipped before a check occurs that the unit can actually talk to the village, only the code doesn't check for the presence of any other unit that could talk to the chief within the selection of units.

raphaelmonticello:
`speakWithChief` on a mixed selection could pick `isNoBadGoodies` even when that unit cannot talk (treasure, wagon, goods).

A treasure listed ahead of a failed trader then did nothing when you clicked talk to the chief.

Only units that `canSpeakWithChief` are considered. A scout in the same selection is still preferred.

Needs a rebuilt DLL.

Fixes #1030

### DIFF
--- a/Project Files/DLLSources/CvDLLButtonPopup.cpp
+++ b/Project Files/DLLSources/CvDLLButtonPopup.cpp
@@ -3904,7 +3904,7 @@ bool CvDLLButtonPopup::launchTalkNativesPopup(CvPopup* pPopup, CvPopupInfo& info
 	gDLL->getInterfaceIFace()->popupSetBodyString(pPopup, gDLL->getText("TXT_KEY_TALK_NATIVES_POPUP", pCity->getNameKey()));
 
 	int iNumActions = 0;
-	if (pUnit->canSpeakWithChief(pUnit->plot()))
+	if (pUnit->canDoCommand(COMMAND_SPEAK_WITH_CHIEF, -1, -1, false, false))
 	{
 		++iNumActions;
 		gDLL->getInterfaceIFace()->popupAddGenericButton(pPopup, gDLL->getText("TXT_KEY_TALK_NATIVES_POPUP_CHIEF"), NULL, COMMAND_SPEAK_WITH_CHIEF);
--- a/Project Files/DLLSources/CvSelectionGroup.cpp
+++ b/Project Files/DLLSources/CvSelectionGroup.cpp
@@ -4474,16 +4474,11 @@ void CvSelectionGroup::speakWithChief()
 	{
 		CvUnit* pLoopUnit = ::getUnit(pUnitNode->m_data);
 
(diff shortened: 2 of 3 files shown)


Schmiddie1979:
I tested this PR and it seems to fix only part of the problem.

With a mixed selection, the "Talk to Chief" button is now displayed, but it is still disabled and cannot be clicked. This means that the new logic in `CvSelectionGroup::speakWithChief()` is apparently never reached in this case.

It looks like the problem occurs earlier, when the game determines whether `COMMAND_SPEAK_WITH_CHIEF` can actually be executed. `CvUnit::canDoCommand()` checks `canSpeakWithChief(plot())` for an individual unit. If the relevant unit/group head of the selection is, for example, a Treasure, this returns false even though another unit in the same selection would be able to speak with the Chief.

So there seems to be an inconsistency between the selection/group command check and the new execution logic:

* The button is displayed because the selection contains a unit that could speak with the Chief.
* The button is disabled because the command check apparently evaluates a unit that cannot speak with the Chief.
* Therefore `CvSelectionGroup::speakWithChief()` is never called, and the new selection of a valid `pBestUnit` cannot solve the problem.

There is also another strange effect: after this happens, selecting the unit that should be able to speak with the Chief individually still does not work. Only after another suitable unit arrives does speaking with the Chief become possible again. This suggests that there may also be some group/selection state involved.

I think the command check for `COMMAND_SPEAK_WITH_CHIEF` needs to be examined as well. For a group selection, the command should be executable if at least one unit in the group passes `canSpeakWithChief()`. Then `CvSelectionGroup::speakWithChief()` can select the best valid unit, with the Scout still taking priority.

So I would not merge this yet. The change in `speakWithChief()` itself looks reasonable, but the preceding command/selection check still seems to prevent it from working correctly.

raphaelmonticello:
Yes. `speakWithChief()` was only the execution side. The exe enables the action from the head unit, so a treasure at the front of the selection left Talk to Chief disabled even when a failed trader in the same group could speak.

Pushed a follow-up: `canDoCommand(COMMAND_SPEAK_WITH_CHIEF)` is true if any unit in the group can speak. The natives popup uses the same check, and the command always dispatches to the group so the scout/failed trader is the one who talks. Needs a rebuilt DLL.

Schmiddie1979:
thank you for the fix

PR #1305, opened 2026-08-30
Show open town jobs on the graduation popup

raphaelmonticello:
The graduation popup listed expert names only. You had to open each town to see if a carpenter shop or fish plot was actually empty.

Each specialty now shows how many open jobs that expert can fill across your towns:

- Workshop professions (carpenter, weaver, ...): unused slots in that special building.
- Plot professions (farmer, fisherman, ...): unused city plots that produce that yield.

Example: `Master Carpenter (3)` means three empty carpenter-shop slots. Tuition gold still shows after that when the specialty costs money.

Needs a rebuilt DLL.

### DIFF
--- a/Project Files/DLLSources/CvDLLButtonPopup.cpp
+++ b/Project Files/DLLSources/CvDLLButtonPopup.cpp
@@ -2194,6 +2194,86 @@ bool CvDLLButtonPopup::launchChooseYieldBuildPopup(CvPopup* pPopup, CvPopupInfo 
 	return (true);
 }
 
+static ProfessionTypes getProfessionForEducatedUnit(UnitTypes eUnit)
+{
+	const UnitClassTypes eUnitClass = GC.getUnitInfo(eUnit).getUnitClassType();
+	for (ProfessionTypes eProfession = FIRST_PROFESSION; eProfession < NUM_PROFESSION_TYPES; ++eProfession)
+	{
+		const CvProfessionInfo& kProfession = GC.getProfessionInfo(eProfession);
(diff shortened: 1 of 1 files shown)


raphaelmonticello:
Requested here:  https://discord.com/channels/788850075673690112/1526021843088314458
It's a simple and elegant solution for Player QoL

Schmiddie1979:
Result from a short test: Could you please check the code. It does currently only show open slots but does not consider if plots are used by units which are no specialists for this plot, e.g. I had one normal colonist and one carpenter in the carpenters shop - hence one slow would be available for a specialist - the overview shows 0 open slots. Would be cool if you could change that. 

What I would particularly also like to verify is how the calculation currently works for buildings that support multiple professions sharing the same workplace slots.

raphaelmonticello:
Yes. The first version counted only physically empty slots, so a colonist already sitting in the carpenter shop made Master Carpenter show 0.

Pushed a follow-up: open jobs are empty slots plus slots/plots worked by a non-expert. A free colonist in the shop is an opening for the specialist. Shared workplaces (several butcher professions, one meat building) still count another specialist in that building as filling a slot. Needs a rebuilt DLL.

Schmiddie1979:
Its getting better but still not correct :-)

For food worker all slots seem to be counted which are not used by an expert - irrespective if fisher or farmer:

image
image

Schmiddie1979:
I made a suggestion for outdoor worker, please check if you like it.

raphaelmonticello:
Looks good. Farmer and fisherman both produce food, so the old plot loop counted the same empty food tiles for both. Filtering land vs water with kProfession.isWater() is the right split.

Merge whenever you like.

PR #1310, opened 2026-09-04
Name occupied city workshops on hover, not just their yield.

raphaelmonticello:
The production box, yield icon, and indoor workers sit on top of the building graphic, so hovering a potter shop showed Pottery prices instead of the building name. Custom House had no overlay, so it already named itself. Point those overlays at building help and name the workshop first on indoor citizen hover.

### DIFF
--- a/Assets/Python/Screens/CvMainInterface.py
+++ b/Assets/Python/Screens/CvMainInterface.py
@@ -1127,7 +1127,9 @@ class CvMainInterface:
 						break
 
 			screen.setImageButton("CityBuildingGraphic" + str(iSpecial), "", BUILDING_GRID[iSpecial][0] + STACK_BAR_HEIGHT, BUILDING_GRID[iSpecial][1] + CITY_TITLE_BAR_HEIGHT, BUILDING_GRID[iSpecial][3], BUILDING_GRID[iSpecial][3], WidgetTypes.WIDGET_ASSIGN_CITIZEN_TO_BUILDING, iSpecial, iTargetBuilding)
-			screen.addDDSGFC("ProductionBox" + str(iSpecial), ArtFileMgr.getInterfaceArtInfo("INTERFACE_PRODUCTION_BOX").getPath(), BUILDING_GRID[iSpecial][0] + STACK_BAR_HEIGHT, BUILDING_GRID[iSpecial][1] + BUILDING_GRID[iSpecial][2], (BUILDING_GRID[iSpecial][3] + STACK_BAR_HEIGHT)*3/4, STACK_BAR_HEIGHT*3/4, WidgetTypes.WIDGET_GENERAL, -1, -1)
+			# Occupied workshops draw this box and the yield icon on top of the building.
+			# Keep the building-help widget so hover still names the building, like Custom House.
+			screen.addDDSGFC("ProductionBox" + str(iSpecial), ArtFileMgr.getInterfaceArtInfo("INTERFACE_PRODUCTION_BOX").getPath(), BUILDING_GRID[iSpecial][0] + STACK_BAR_HEIGHT, BUILDING_GRID[iSpecial][1] + BUILDING_GRID[iSpecial][2], (BUILDING_GRID[iSpecial][3] + STACK_BAR_HEIGHT)*3/4, STACK_BAR_HEIGHT*3/4, WidgetTypes.WIDGET_ASSIGN_CITIZEN_TO_BUILDING, iSpecial, iTargetBuilding)
 			screen.hide("ProductionBox" + str(iSpecial))
 
 	# BUTTONS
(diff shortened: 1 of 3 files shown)


raphaelmonticello:
issue: 
image

raphaelmonticello:
Issue raised by Tish

Schmiddie1979:
I think this should be examined much more closely before treating it as a bug and changing the existing UI behavior.

As I understand the current city screen design, the different hover areas intentionally provide different kinds of information:

* Hovering over the building itself provides information about the building.
* Hovering over a citizen working inside the building provides information about that citizen, profession and production.
* Hovering over the yield/production display provides information about the produced yield.

Indoor workers have always been displayed on top of the building graphics. This is not an exceptional situation affecting only certain workshops, but a fundamental part of the city screen layout. Every occupied workshop has its citizens displayed over parts of the building graphic.

For that reason, I do not think that the fact that a citizen covers part of the building graphic should in itself be considered a bug. If I hover over a citizen, I actually expect to see information about that citizen. Likewise, if I hover over the production/yield display, I expect to see the corresponding yield information. The building can still be identified by hovering over the remaining building area, especially the upper part of the building graphic.

There is another important concern with this PR.

I have spent considerable effort improving the production and workshop tooltips. These tooltips provide much more than just the name of the produced yield. They are intended to give the player a useful production overview, including detailed information about inputs and outputs and, where applicable, market and price information.

This PR changes several UI elements from the existing yield-help widgets (`WIDGET_HELP_YIELD` / `WIDGET_HELP_TWO_YIELDS`) to `WIDGET_ASSIGN_CITIZEN_TO_BUILDING`. That is not merely adding the building name to an existing tooltip. It changes which type of tooltip is requested for these UI areas.

Therefore I am concerned that solving this supposed display issue could unintentionally remove or hide parts of the production information that I deliberately added to these hover areas. In other words, useful and carefully developed functionality might be lost in order to solve something that may actually just be the intended city screen layout.

The same applies to the citizen tooltip. I do not think that the building name necessarily needs to be inserted simply because the citizen graphic happens to overlap the building. The citizen is a separate UI element, and when hovering over it the primary information should remain information about the citizen.

So before making this change, I think a concrete case should first be established where the current behavior is actually broken. For example:

1. Is there an occupied building which cannot be identified anywhere by hovering over the building itself?
2. Which information is currently available when hovering over the building, the citizen, the production number and the yield icon respectively?
3. Which of this information disappears or changes with this PR?
4. Are the detailed production/yield tooltips still fully available after the change?
5. Does this interact with the recently improved workshop production overview and its market/price information?
6. How do buildings with several input/output yields behave?
7. Is there actually missing information, or is this simply the intended separation between building, citizen and yield hover areas?

My current view is therefore that the existing behavior should not be changed merely because workers overlap the building graphics. That overlap is normal and intentional in the city screen and does not by itself constitute a bug.

If there is a specific building where no usable building-hover area remains, I think that particular case should be investigated. But I would strongly prefer not to redirect the existing citizen/yield hover areas globally unless it has first been verified that this is necessary and that none of the existing production information is lost.

At minimum, this needs much more detailed testing before the PR can be considered safe. In its current form I would not regard the described behavior as a bug that automatically needs to be fixed.

Schmiddie1979:
I will review it, please wait.

Schmiddie1979:
Thank you very much for taking the time to investigate this issue and for preparing the PR. I appreciate the effort and the intention behind it.

Unfortunately, after testing the complete PR and looking more closely into the original issue, **I have to reject it.**

The main problem is that the changes interfere with the production/yield tooltips we have built up over time regarding the production and domestic market information. These tooltips deliberately provide quite a lot of useful information about production, yields, market prices, domestic demand, etc. With this PR, parts of that functionality are replaced by building information. From my perspective, losing this carefully developed production overview would be a significant regression.

More importantly, it turns out that these code changes **are not actually necessary to solve the underlying issue.** The Pottery behaves differently only because of what appears to be a UI/display initialization problem: directly after starting the game, part of the building graphic does not provide the expected building tooltip. However, after switching out of the game and back into it, the hover works correctly without any further change. This strongly suggests that the actual problem is related to the initial display or widget ordering rather than to the tooltip logic itself.

The building information for the Pottery is also already accessible from the beginning by hovering over the part of the building below the production bar. So the building tooltip itself is present and functional; only part of its hover area is initially covered or intercepted.

Finally, all the other buildings work correctly. This is therefore a very limited display issue affecting the Pottery rather than a general problem with the city-screen tooltip design.

For these reasons, I don't think the general hover behavior of the city screen should be changed to work around this one display issue, especially when doing so removes useful information from the production interface.

So, although I very much appreciate the work and the attempt to improve this, I have to reject and close this PR. I would rather keep the existing production/yield information and accept this minor Pottery display issue than introduce a broader UI regression.

Thank you again for taking the time to work on it.

Schmiddie1979:
Nevertheless I will further investigate what happens and why - challenge accepted. There must be a difference in one of the files. Don't know yet what happens. Maybe graphic
 file is different than the other.

